PR #1, opened 2017-09-12
DRAFT: Jenkins Enhancement Proposal

rtyler:
NOTE: This should be squashed when merged into `master`


rtyler:
@jenkinsci/jep-editors If you could spare some time to look over this format and make sure that it's clear and concise. Assuming the document is clear, I wish to send this to the jenkinsci-dev@ list before the end of the week to start building consensus around the approach.

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+Each JEP should have the following parts:
+
+. **Metadata** - table containing metadata about the JEP, including the JEP
+  number, a short descriptive title, the names, and optionally the contact info
+  for each author, etc.
+. **Abstract** - short (200 word) description of the technical issue
+  being addressed.
+. **Specification** - The technical specification should describe the
+  syntax and semantics of any new feature. The specification should be
+  sufficiently detailed to allow new, or existing, Jenkins developers to
```
omehegan: You can eliminate the commas around "or existing" in this context.

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
++
+The rationale should provide evidence of consensus within the
+community and discuss important objections or concerns raised
+during discussion.
+
+. **Backwards Compatibility** - All JEPs that introduce backwards
+  incompatibilities must include a section describing these
+  incompatibilities and their severity. The JEP must explain how the
+  author proposes to deal with these incompatibilities. JEP
+  submissions without a sufficient backwards compatibility treatise
```
omehegan: I think "treatise" is too strong a word here; it's sort of like "manifesto." I would say, "JEP submissions that do not adequately discuss backwards compatibility, when such discussion is required, may be rejected outright."

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+  may be rejected outright.
+. **Reference Implementation** -- The reference implementation must be
+  completed before any JEP is given status "Final", but it need not
```
omehegan: "Final,"

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+. **Status** -- Draft :speech_balloon:, Deferred :hourglass:, Accepted :ok_hand:, Rejected :no_entry:, Withdrawn :hand:, Final :lock:, Replaced :dagger:, Active :smile:.
+. **Type** -- Describes the type of JEP: Standards, Informational, Process
+. **Created** -- Date (`%Y%m%d`) when the document was first created.
+
+
+
+A **BDFL-Delegate** row is used to record cases where the final decision to
+approve or reject a JEP rests with someone other than the BDFL. (The
+delegate's email address is currently omitted due to a limitation in the
+email address masking for reStructuredText JEPs)
```
omehegan: "reStructuredText JEPs.)"

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+=== Auxiliary Files
+
+JEPs may include auxiliary files such as diagrams. Such files must be
+named appropriately, with lowercase letters and no spaces, and be included in
+the directory with the `README.adoc` describing the JEP.
+
+
+=== Reporting JEP Bugs, or Submitting JEP Updates
+
+How you report a bug, or submit a JEP update depends on several factors, such
```
omehegan: Drop comma after "bug"

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+original author no longer has the time or interest in updating it or
+following through with the JEP process, or has fallen off the face of
+the 'net (i.e. is unreachable or not responding to email). A bad
+reason to transfer ownership is because the author doesn't agree with the
+direction of the JEP. One aim of the JEP process is to try to build
+consensus around a JEP, but if that's not possible, an author can always
+submit a competing JEP.
+
+If you are interested in assuming ownership of a JEP, you can also do this via
+pull request. Fork the `JEP repository`_, make your ownership modification, and
```
omehegan: Funny syntax around "JEP repository?"

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+
+=== JEP Editor Responsibilities & Workflow
+
+A JEP editor must subscribe to the jenkinsci-dev@googlegroups.com list and must
+watch the JEP repository footnoteref:[repo]. Most correspondence regarding JEP
+administration can be handled through GitHub issues and pull requests.
+
+For each new JEP that comes in an editor does the following:
+
+* Read the JEP to check if it is ready: sound and complete. The ideas
```
omehegan: Maybe, "ready, sound, and complete."

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+JEP review and resolution may also occur on a list other than jenkinsci-dev@ In
+this case, the "Discussions-To" heading in the JEP will identify the
+appropriate alternative list where discussion, review and pronouncement on the
+JEP will occur.
+
+For a JEP to be accepted it must meet certain minimum criteria:
+
+* It must be a clear and complete description of the proposed enhancement.
+* The enhancement must represent a net improvement.
+* The proposed implementation, if applicable, must be solid and must not complicate Jenkins unduly.
```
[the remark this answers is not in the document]
rtyler: PEP-1 includes some subjective text purposefully to allow bending of the interpretation later down the line.

I'm partial to some declaration of "the proposed implementation shouldn't be a garbage fire"

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+
+Once a JEP has been accepted, the reference implementation must be
+completed. When the reference implementation is complete and incorporated
```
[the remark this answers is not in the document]
rtyler: In the case of a Process JEP, a reference implementation might incorporate some workflow changes, or something like that.

In general, the "reference implementation" might be better phrased here as "the implementation." As "code talks" being a central tenet of the project, I'll tidy this a bit and clarify

jep/1/README.adoc
```diff
@@ -0,0 +1,536 @@
...
+A JEP can also be "Rejected". Perhaps after all is said and done it
+was not a good idea. It is still important to have a record of this
+fact. The "Withdrawn" status is similar - it means that the JEP author
+themselves has decided that the JEP is actually a bad idea, or has
+accepted that a competing proposal is a better alternative.
+
+When a JEP is Accepted, Rejected or Withdrawn, the JEP should be updated
+accordingly. In addition to updating the status field, at the very least
+the Resolution header should be added with a link to the relevant post
+in the python-dev mailing list archives.
```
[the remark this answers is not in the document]
rtyler: copy paste fail :smiling_imp:

rtyler:
@omehegan thanks for the thorough review! I've responded to, or implemented, the changes you've suggested.

jep/1/README.adoc
daniel-beck: Upcoming JEP: Reject this grammar abomination that only results in confusion internationally for future JEPs.

omehegan: Sorry, it's just American style. At first I thought it was a one-off error in the doc, but I guess Tyler used Euro style throughout. I don't care which convention is used as long as it's consistent.

daniel-beck: I know this rule exists, and makes for better prose reading with direct speech, but this doesn't make sense for JEPs and similar documentation.

Right now, it's inconsistent: 
https://github.com/jenkinsci/jep/pull/1/files#diff-43894e615b82e2536b1341c33ff49afeR86
https://github.com/jenkinsci/jep/pull/1/files#diff-43894e615b82e2536b1341c33ff49afeR158
https://github.com/jenkinsci/jep/pull/1/files#diff-43894e615b82e2536b1341c33ff49afeR250

rtyler:
Can I get a final ruling on how to use quotes and punctuation plz.

jep/1/README.adoc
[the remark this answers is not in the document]
omehegan: OK, go with it then.

jep/1/README.adoc
omehegan: It's not a hill I'm prepared to die on, if you feel like it should be the other way I'm not going to object 👍

omehegan:
Accidentally started another review. I defer to your judgment, neither style is obscuring the meaning IMO. http://www.thepunctuationguide.com/british-versus-american-style.html for reference. I simply think punctuation outside of quotes will read weirdly to any North American reader, hence my impulse to change it. But I'm not here to be the grammar police :)

rtyler:
@omehegan Okay, I prefer the punctuation outside the quotes, so I'm just going to make this consistent on that style for this document at least. I honestly don't care about defining a styleguide for this kind of thing for JEPs :smile:

rtyler:
I'm going to squash all these commits down to just one commit to kick off the discussion on the JEP

jep/1/README.adoc
```diff
@@ -0,0 +1,542 @@
...
+
+image::workflow.png[JEP Workflow]
+
+Some Informational and Process JEPs may also have a status of "Active" if they
+are never meant to be completed. E.g. JEP 1 (this JEP).
+
+
+=== JEP Maintenance
+
+In general, Standards track JEPs are no longer modified after they have
```
bitwiseman: are no longer modified -> are not modified

jep/1/README.adoc
```diff
@@ -0,0 +1,542 @@
...
+ Standards Track JEPs but apply to areas other than the Jenkins codebase
+ itself. They may propose an implementation, but not for what would be
+ generally considered the Jenkins codebase; they often require community
+ consensus; unlike Informational JEPs, they are more than recommendations, and
+ users are typically not free to ignore them. Examples include procedures,
+ guidelines, changes to the decision-making process, and changes to the tools
+ or environment used in Jenkins development. Any meta-JEP (JEP focusing on the
+ improvement of the Jenkins Enhancement Proposal process) is also considered a
+ Process JEP.
+
```
bitwiseman: Suggest creating a `== JEP Terminology` here.  and move "Jenkins's BDFL" and "JEP Editors" under it.

jep/1/README.adoc
```diff
@@ -0,0 +1,542 @@
...
+
```
bitwiseman: Need a section that defines one or more set of terms for "Must/Always, Should, May/Can, Should not, Never/Must Not".   

For example, Microsoft .NET Framework Guidelines use "Do, Consider, Avoid, Do Not", have special formatting for list items using those words: 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces

I'm not married to any particular set, but I think it is important to lay out terms for all future JEPs and have a clear structure that highlights intent and requirements.

jep/1/README.adoc
```diff
@@ -0,0 +1,542 @@
...
+== What belongs in a successful JEP?
+
+Each JEP should have the following parts:
+
+. **Metadata** - table containing metadata about the JEP, including the JEP
+  number, a short descriptive title, the names, and optionally the contact info
+  for each author, etc.
+. **Abstract** - short (200 word) description of the technical issue
+  being addressed.
+. **Specification** - The technical specification should describe the
```
bitwiseman: Specification should come after motivation.

jep/1/README.adoc
```diff
@@ -0,0 +1,542 @@
...
+. Squash the commit(s) into the branch.
+. Update the JEP number in the document.
+
+Updates to existing JEPs should be submitted as a GitHub pull request.
+
+JEP editors don't pass judgment on JEPs. They merely do the
+administrative & editorial part (which is generally a low volume task).
+
+
+== Rationale
```
bitwiseman: This sounds more like a "Motivation" that a "Rationale".

rtyler:
@bitwiseman I agree with a number of your proposed changes. Pulling key points out of the "mush" worries me because I think this process is important enough to not be abbreviated, or "tl;dr'ed".

I'll create a pull request against the `jep-1` branch with some restructurings.

rtyler:
Oh, also, if GitHub would properly render the AsciiDoc `:toc:` macro, this would be such much more readable :man_facepalming:

jep/1/README.adoc
rtyler: This comment requires more explanation if we're to make this change. This is the ordering copied from PEP-1, and I personally find Abstract/Specification to be a reasonable first couple sections to review of a technical design document.

bitwiseman: @rtyler Thinking about it a bit more, I agree.  Putting the Motivation first helps set context for those unfamiliar with it.  But Motivation is fluff for those already aware of the it.

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+  describing what motivated the design and why particular design
+  decisions were made. It should describe alternate designs that
+  were considered and related work, e.g. how the feature is supported
+  in other languages.
++
+The rationale should provide evidence of consensus within the
+community and discuss important objections or concerns raised
+during discussion.
+
+. **Backwards Compatibility** - All JEPs that introduce backwards
```
batmat: Are all parts *required* in a JEP? I mean: is anyone **required** to have a `Backwards Compatibility` part, and possibly put `N/A` on it? I feel the latter would be better for clarity.

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+The final implementation must include test code and documentation
+appropriate for either the Jenkins user or developer documentation.
+
+
+==== JEP Formats and Templates
+
+JEPs are UTF-8 encoded text files using the
+link:https://asciidoctor.org[AsciiDoc] format.  AsciiDoc allows for rich markup
+that is still quite easy to read, but also results in good-looking and
+functional HTML.
```
batmat: Probably my comment above about "one line per sentence for diff-friendliness" would be in that section, if so.

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+| Type
+| Process
+
+| Created
+| 2017-09-12
+|===
+----
+
+
+. **JEP** -- Proposal number, given by the JEP editors. Use "9999" until one is assigned.
```
batmat: Stupid question: but let's assume this process gets great traction, why not `0000` instead of `9999` that we _may_ reach some day. (OK, maybe you actually want to do that because it would show a great success, "a nice problem to have" ;-))

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+rendered obsolete by a later document; the value is the number of the JEP that
+replaces the current document. The newer JEP must have a **Replaces** row
+containing the number of the JEP that it rendered obsolete.
+
+
+==== Auxiliary Files
+
+JEPs may include auxiliary files such as diagrams. Such files must be
+named appropriately, with lowercase letters and no spaces, and be included in
+the directory with the `README.adoc` describing the JEP.
```
batmat: Having myself pushed big garbage in the past, I feel something like "a particular attention to the reasonable size of the resources and resolution will be made".
IOW, a diagram that could be readable in a 5x10cm rectangle *must* not be 12 MB.
Because this will impact the size and the easiness of cloning it then forever.

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+revision, with specific instructions.
+
+Once the JEP is ready for the repository, a JEP editor will:
+
+. Assign a JEP number (almost always just the next available number, but
+  sometimes it's a special/joke number, like 666 or 3141).
+. Create a new branch for the JEP, i.e. `jep-1`.
+. Retarget the original author's pull request to the new branch
+. Squash the commit(s) into the branch.
+. Update the JEP number in the document.
```
batmat: OK I think that clarifies my question above. So we'd have as many branches as there JEPs, forever, right?

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+below:
+
+. **MUST** This word, or the terms "REQUIRED" or "SHALL", mean that the
+   definition is an absolute requirement of the specification.
+. **MUST NOT** This phrase, or the phrase "SHALL NOT", mean that the
+   definition is an absolute prohibition of the specification.
+. **SHOULD** This word, or the adjective "RECOMMENDED", mean that there
+   may exist valid reasons in particular circumstances to ignore a
+   particular item, but the full implications must be understood and
+   carefully weighed before choosing a different course.
```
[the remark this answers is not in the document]
bitwiseman: @batmat - This is verbatim quote from the linked document for convenience and clarity.  
I'll add a NOTE at the end of this list the the effect that when going counter to "Should" or "Should not" guidance, one SHOULD provide and explanation. 😃

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+. **SHOULD NOT** This phrase, or the phrase "NOT RECOMMENDED" mean that
+   there may exist valid reasons in particular circumstances when the
+   particular behavior is acceptable or even useful, but the full
+   implications should be understood and the case carefully weighed
```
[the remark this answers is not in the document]
bitwiseman: same. 😄

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
...
+   before implementing any behavior described with this label.
+. **MAY** This word, or the adjective "OPTIONAL", mean that an item is
```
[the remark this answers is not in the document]
bitwiseman: same. 😄

jep/1/README.adoc
bitwiseman: I like `0000` as the marker.

jep/1/README.adoc
bitwiseman: @batmat  Could you submit a PR that includes you suggested changes?  What balance would you like to strike?

bitwiseman:
@rtyler - Once #7 is merged, I'm calling done making changes.  If accepted, I'll create the `jep-submission` branch.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+for the JEP repository footnoteref:[repo].
+
+The JEP editors will not unreasonably deny a JEP.
+Reasons for denying JEP "Draft" status include:
+
+* duplication of effort
+* being technically unsound
+* not providing enough information in the
+  "Motivation" or "Backwards Compatibility" sections
+* not in keeping with the Jenkins philosophy.
```
oleg-nenashev: It would be great to describe it somewhere. I am pretty sure everyone has different understanding of Jenkins philosophy

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+the sumbmission PR will have been merged to a matching folder and feature branch
+(i.e the
+`link:https://github.com/jenkinsci/jep/tree/jep-1/jep/1[jep/1]` folder in the
+`https://github.com/jenkinsci/jep/tree/jep-1/[jep-1]` branch).
+
+IMPORTANT: "Approval as Draft" is *not* the same as accepting the JEP.
+
+Editors are not the only ones who can approve as submission.
+Non-editor contributors who have "git push" privileges for the
+JEP repository footnoteref:[repo] may also approve submissions.
```
oleg-nenashev: It means jenkinsci GitHub organization admins, I'd guess.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+as ideas that sound good in principle sometimes turn out to be impractical
+when subjected to the test of implementation.
+
+A JEP's author is responsible for collecting community feedback on a JEP
+before submitting it for review.
+Potential changes to a draft JEP may be discussed further on jenkinsci-dev@.
+However, long open-ended discussions are not recommended on mailing lists.
+Strategies to keep the discussion efficient include:
+
+* setting up a series of in-person, or video-conferencing sessions to
```
oleg-nenashev: Maybe we could revive office hours timeslots for that. Having a regular timeslot is useful imho

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+  discuss the JEP with necessary stakeholders.
+* having the JEP author accept private comments in the early design phases
+* setting up a wiki page, etc.
+* committing and reviewing small concrete changes via Pull Requests
+  rather than large sweeping changes
+
+JEP authors should use their discretion here.
+
+The JEP author may also ask JEP editors for further feedback regarding the
+style and consistency of a JEP and its readiness for review by the BDFL.
```
oleg-nenashev: s/BDFL/Governance meeting/

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+
+As updates are necessary, the JEP author and other contributors
+should push commits to their fork of the JEP repository footnoteref:[repo],
+and submit pull requests targeting the JEP's feature branch.
+
+[[review]]
+==== JEP Review
```
oleg-nenashev: Whatever we select as a final review stage (BDFL/Governance Meeting/...), there must be SLAs, at least for the first reply and for the decision. Otherwise the proposals may be killed by ignorance as it sometimes happens in the mailing list now

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+JEP will occur.
+
+[[accepted]]
+==== Accepting a JEP
+
+For a JEP to be "Accepted" it must meet certain minimum criteria:
+
+* It must be a clear and complete description of the proposed enhancement.
+* The enhancement must represent a net improvement.
+* The proposed implementation, if applicable, must be solid and must not complicate Jenkins unduly.
```
oleg-nenashev: ADD: The proposed implementation must be open-source

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+This is intended for Informational JEPs, where version 2 of an API can replace version 1.
+
+==== Other JEP Outcomes
+
+Not all JEPs will be accepted and finalized.
+
+Rejected::  [[rejected]]
+A JEP can also be "Rejected". Perhaps after all is said and done it
+was not a good idea. It is still important to have a record of this
+fact.
```
oleg-nenashev: Would be great to have an advice to the author about next steps. E.g. how to escalate or resubmit proposal

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+==== Updating JEP Status and Resolution
+
+Whenever a JEP status changes, the "Status" field in the JEP document must be updated.
+
+The possible paths of a JEP's status are as follows:
+
+.JEP Workflow
+image::workflow.png[JEP Workflow]
+
+When a JEP is Accepted, Rejected or Withdrawn,
```
oleg-nenashev: Not sure what it means. Maybe a mirror JIRA issue in the "JEP" project is preferable

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+  reasonably understand the scope/impact of an implementation.
+. **Motivation** - The motivation is critical for JEPs that want to
+  change Jenkins itself. It should clearly explain why the
+  existing code base is inadequate to address the
+  problem that the JEP solves. JEP submissions without sufficient
+  motivation may be rejected outright.
+. **Reasoning** - The reasoning describes why particular design
+  decisions were made. It should describe alternate designs that
+  were considered and related work, e.g. how the feature is supported
+  in other languages.
```
oleg-nenashev: "languages" comes from Python. In the case of Jenkins I would rather propose using generic "systems" or "solutions"

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+  being addressed.
+. **Specification** - The technical specification should describe the
+  syntax and semantics of any new feature. The specification should be
+  sufficiently detailed to allow new or existing Jenkins developers to
+  reasonably understand the scope/impact of an implementation.
+. **Motivation** - The motivation is critical for JEPs that want to
+  change Jenkins itself. It should clearly explain why the
+  existing code base is inadequate to address the
+  problem that the JEP solves. JEP submissions without sufficient
+  motivation may be rejected outright.
```
oleg-nenashev: Should be rejected during the discussion in the ML imho

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+community and discuss important objections or concerns raised
+during discussion.
+
+. **Backwards Compatibility** - All JEPs that introduce backwards
+  incompatibilities must include a section describing these
+  incompatibilities and their severity. The JEP must explain how the
+  author proposes to deal with these incompatibilities. JEP submissions that do
+  not adequately discuss backwards compatibility, when such discussion is
+  required, may be rejected outright.
+. **Reference Implementation** -- The reference implementation must be
```
oleg-nenashev: "open-source reference implementation"?

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+
+The final implementation must include test code and documentation
+appropriate for either the Jenkins user or developer documentation.
+
+
+==== JEP Formats and Templates
+
+JEPs are UTF-8 encoded text files using the
+link:https://asciidoctor.org[AsciiDoc] format.
+AsciiDoc allows for rich markup that is still quite easy to read,
```
oleg-nenashev: Tables are bad in any markup language, see this doc :(
For design docs I use many tables, but it's something to deal with

jep/1/README.adoc
oleg-nenashev: UPD: "ignorance" means that it slips in priority, gets missed, etc.

jep/1/README.adoc
daniel-beck: Too weak if anything about this should be added. Same license as the component it's meant to be added to, or MIT if something new, perhaps?

oleg-nenashev: Works for me

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+NOTE: Small enhancements or patches often don't need a JEP can be handle via the Jenkins
+development workflow with a JIRA item and/or pull request to the appropriate repository.
+
+==== Find a champion
+
+Each JEP must have a champion -- someone who writes the JEP using the style and
+format described below, shepherds the discussions in the appropriate forums,
+and attempts to build community consensus around the idea. The JEP champion
+(a.k.a. <<Author>>) should first attempt to ascertain whether the idea is JEP-able.
+Posting to the jenkinsci-dev@googlegroups.com mailing list is the best way to
```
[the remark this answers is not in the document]
rtyler: I would like to keep the default here to `jenkinsci-dev@` for now to reduce fragmentation of discussion and make it clearer to contributors where discussions should start.

This is changeable later, but I think it's important to keep it simple for now.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+gives the author a chance to flesh out the draft JEP to make sure it is
+properly formatted, of high quality, and to address initial concerns about the
+proposal.
+
+[[submission]]
+==== Creating a JEP Submission
+
+Following a discussion on jenkinsci-dev@,
+the proposal should be turned into as a JEP submission and submitted
+via a GitHub pull request to this repository footnoteref:[repo].
```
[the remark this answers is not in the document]
rtyler: > This step likely means "Convert the document from a reviewable format (e.g. Google Doc) to Asciidoc". It is pretty trivial, but it still requires extra efforts. Assuming that an abstract company does internal reviews before coming up with a proposal, it means that there are likely 2 conversion steps for JEPs.

I interpret this to mean "CloudBees will have to do two conversion steps." Which I have no problems with whatsoever :smile: 

> I am not sure it's friendly to individual contributors. So such JEP process IMHO should not be required for 90% of small- and medium-scale enhancements in the core.

I definitely agree that small/medium-scale enhancements shouldn't necessarily need JEPs, but this process relies on contributor's using their good judgement to determine the scope and impact of what they're trying to accomplish, and for other contributors to helpfully guide them down a path (small -> PR, non-small impact -> JEP).

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+. Create a new branch called `jep-submission` their fork
+  (`git checkout -b jep-submission`).
+. Copy the folder `jep-template/0000` to `jep/0000`.
+. Modify the template JEP in `jep/0000`
+  per the instructions in this JEP (which are also outlined in the template).
+. Commit and push the changes to their fork
+  and submit a pull request targeting the `jenkinsci/master` branch.
+
+A JEP editor will check the submission for conformance with
+JEP structure and formatting guidelines.
```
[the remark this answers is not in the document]
rtyler: I've looked for tools to do this for jenkins.io, there's not anything I found which was really good in the "linting" space for AsciiDoc :frowning: 

(disclaimer: I haven't looked again in about 6-8 months)

jep/1/README.adoc
rtyler: Not sure what an "SLA" by volunteer contributors would look like. If nobody replies then it's automatically approved? :smiling_imp:

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+ or environment used in Jenkins development. Any meta-JEP (JEP focusing on the
+ improvement of the Jenkins Enhancement Proposal process) is also considered a
+ Process JEP.
+
+=== JEP Terminology
+
+==== BDFL
+
+There are several references in this JEP to the "BDFL". This acronym stands for
+"Benevolent Dictator for Life", and refers to
```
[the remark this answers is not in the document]
oleg-nenashev: > I will remind you that despite what some contributors seem to think, in the US "KKK" refers to a hateful group of racists and domestic terrorists.

Well, this is my turn to apologize then. My only excuse is that I am not an American citizen/resident and not a native speaker. 

Regarding this group, I know Ku Klux Klan only by the full name, and actually I have not really thought that "Klan" starts from "K" instead of "C". So for me this abbreviation never meant this group which I definitely do not support. And I definitely do not find this abbreviation funny after the clarification (probably I have missed the previous one).

I will stop using this abbreviation for sure. Unfortunately such mistakes happen too often in a global world, especially in the case case of acronyms.

jep/1/README.adoc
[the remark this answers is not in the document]
oleg-nenashev: agreed

jep/1/README.adoc
oleg-nenashev: It could be something like that...

BDFL mode:
1) BDFL should provide initial feedback in 2 weeks and make a decision within 1 month
2) If BDFL does not make a decision, the submitter is eligible to escalate JEP to the governance meeting

Governance meeting mode:
1) The JEP owner is responsible to add JEP discussion to the Governance meeting agenda and to propose JEP reviewers. 
2) Governance meeting approves JEP reviewers or selects other ones, who will act as a steering committee
3) Steering committee should perform review of the proposal (in public) and then reports the result at the governance meeting 1 month later (3 options: approve, reject, another cycle required)
4) If the steering committee recommends approve/reject, the Governance meeting votes for the decision.

jep/1/README.adoc
oleg-nenashev: @rtyler added patch in another PR, I have some concerns about the "same license" requirement. See https://github.com/jenkinsci/jep/pull/8#discussion_r139896232

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+The JEP editors will not unreasonably deny a JEP.
+Reasons for denying JEP "Draft" status include:
+
+* duplication of effort
+* being technically unsound
+* not providing enough information in the
+  "Motivation" or "Backwards Compatibility" sections
+* not in keeping with the Jenkins philosophy.
+
+The BDFL may be consulted during the approval phase,
```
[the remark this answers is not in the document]
oleg-nenashev: Agreed, though it's how our current governance mode works. I second the concern, and probably we should reconsider the governance meeting approach.

jep/1/README.adoc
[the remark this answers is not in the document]
oleg-nenashev: > I interpret this to mean "CloudBees will have to do two conversion steps." Which I have no problems with whatsoever

Well, I  would expect any company-contributor to conduct an internal review before coming up with a JEP-scale proposal. 

> I definitely agree that small/medium-scale enhancements shouldn't necessarily need JEPs, but this process relies on contributor's using their good judgement to determine the scope and impact of what they're trying to accomplish, and for other contributors to helpfully guide them down a path (small -> PR, non-small impact -> JEP).

maybe also "medium -> mailing list". There are several great examples of discussions inside the Jenkins mailing list, sometimes backed by Google Doc proposal. And the difficult thing would be to determine where this approach ends and where JEP starts.

In mailing list an optimistic decision ETA is 1-2 weeks, in JEP it would be rather about 1-2 months from what I see in this document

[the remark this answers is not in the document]
rtyler: > In mailing list an optimistic decision ETA is 1-2 weeks, in JEP it would be rather about 1-2 months from what I see in this document

I think this is overly pessimistic. All of the well-received and well-discussed proposals ([for example](https://groups.google.com/d/msgid/jenkinsci-dev/7a7e2b81-8795-48bd-b1c2-d0ee31123df3%40googlegroups.com)), such as those using Google Docs you mention, very naturally fit into JEP in my opinion.

Creating the document shouldn't be hard, if for some reason it's more difficult to get consensus around a JEP than it is a topic on the mailing list, then I feel like that is an indication of some deeper organizational problem.


Anyways, my preference is to intentionally keep this a _bit_ ambiguous until we figure out (by practicing the process) what distinction for a JEP-able vs. a JEP-unnecessary change makes sense for us.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+* link:https://github.com/oleg-nenashev[Oleg Nenashev]
+* link:https://github.com/bitwiseman[Liam Newman]
+* link:https://github.com/omehegan[Owen Mehegan]
+* link:https://github.com/orrc[Christopher Orr]
+
+JEP editorship is by invitation of the current editors. All of the JEP workflow
+can be conducted via the GitHub JEP Repository footnoteref:[repo] and pull
+requests.
+
+==== Author
```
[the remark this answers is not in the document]
bitwiseman: @oleg-nenashev  I'm going to delete your comment above as duplicate of this one, to centralize this discussion:

 > Rename to <<Champion>>? Author is a legal term, which contradicts with the next sentence (line 65)

I support unifying these terms into one since they are  currently used interchangeably in this document.   @rtyler  Do you have a preference as to which one we use?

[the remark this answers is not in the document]
rtyler: @bitwiseman no, go for what feels most comfortable to you :+1:

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: Adding to Reasoning section.  RESOLVED.

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: resolved?

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: resolved?

jep/1/README.adoc
bitwiseman: @batmat  I believe tyler addressed this with a later change.  Please check and comment "RESOLVED" if resolved.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+repository, their revision history is the historical record of the
+feature proposal footnoteref:[repo, https://github.com/jenkinsci/jep].
+
+
+=== JEP Types
+
+There are three kinds of JEP:
+
+. A **Standards Track** JEP describes a new feature or implementation
+ for Jenkins. It may also describe an interoperability or
```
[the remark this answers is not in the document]
bitwiseman: Adding a ascidoc style guide section.  RESOLVED.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+features, for collecting community input on an issue, and for
+documenting the design decisions that have gone into Jenkins.
+Each JEP has an <<Author>>.
+The author of a JEP may or may not actually write the text of the JEP.
+Regardless, the JEP Author is responsible for building
+consensus for that JEP within the community and
+documenting dissenting opinions.
+
+Because the JEPs are maintained as text files in a versioned
+repository, their revision history is the historical record of the
```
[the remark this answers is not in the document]
bitwiseman: @oleg-nenashev , I think tyler addressed this.  Resolved?

jep/1/README.adoc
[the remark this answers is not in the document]
oleg-nenashev: agreed

jep/1/README.adoc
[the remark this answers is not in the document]
oleg-nenashev: yes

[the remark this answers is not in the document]
oleg-nenashev: thanks @rtyler

jep/1/README.adoc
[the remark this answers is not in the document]
oleg-nenashev: @bitwiseman thanks! Now we probably need to enhance JEP-1 itself :)

jep/1/README.adoc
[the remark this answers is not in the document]
oleg-nenashev: "Sponsor" is quite ambiguous as well, but this term introduced by @bitwiseman solves my original concern about legal clashes. Hence it is RESOLVED

jep/1/README.adoc
```diff
@@ -0,0 +1,767 @@
...
+
+[[accepted]]
+==== Accepting a JEP
+
+For a JEP to be "Accepted" it must meet certain minimum criteria:
+
+* It must be a clear and complete description of the proposed enhancement.
+* The enhancement must represent a net improvement.
+* The proposed implementation, if applicable, must be solid, must not
+  complicate Jenkins unduly, and must be the same license as the component the
```
oleg-nenashev: I am reposting my comment about "same license" in https://github.com/jenkinsci/jep/pull/8#discussion_r139896232, because IMHO it worth discussion:

> "same license" is maybe too strong. But it's better than legal holywars about "compatible licenses".
>
> OTOH some exceptions could be useful. E.g. some companies strictly require "Apache v.2", because MIT is not explicit about the patent release. By setting this condition we explicitly require companies to do reference implementation with the MIT license, and hence such potential contributors may have no way to comply with the process without getting a legal-approved exception (it hurts :( ).
> 
> I would just pre-approve MIT, Apache and BSD licenses for reference implementations, independently of the component license. GPL/LGPL is more complicated in the case of Jenkins.

@daniel-beck @rtyler

daniel-beck: I don't even want to think about the mess that mixed-licensed (i.e. half one license, half the other; not dual licensed!) components and correctly working with that would be. IANAL after all.

rtyler: @oleg-nenashev I would assume most JEPs will refer to core or many of the "essential" plugins, all of which are MIT licensed, such that anybody contributing to those repositories is expected to contribute under the existing (MIT) license.

I believe we should keep this text as-is until we have a reason to change it at some later date.

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: RESOLVED.

jep/1/README.adoc
```diff
@@ -0,0 +1,593 @@
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+= JEP-1: Jenkins Enhancement Proposal Format
```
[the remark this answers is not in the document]
bitwiseman: batmat - So, is this resolved?   No change needed?

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: @oleg-nenashev How do we need to enhance it?  I thought the JEP meets the guidelines as stated.

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+There are three kinds of JEP:
+
+. A **Standards Track** JEP describes a new feature or implementation
+ for Jenkins. It may also describe an interoperability or
+ backwards-compatibility standard which will be supported for the feature in
+ current versions of Jenkins, moving forward.
+. An **Informational** JEP describes a Jenkins design issue, or
+ provides general guidelines or information to the Jenkins community,
+ but does not propose a new feature. Informational JEPs do not
+ necessarily represent a Jenkins community consensus or
```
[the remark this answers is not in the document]
bitwiseman: @oleg-nenashev - Resolved?

jep/1/README.adoc
```diff
@@ -0,0 +1,754 @@
...
+  <<approval, approval as a Draft JEP>>.
+  An editor checks the submission and determines it is ready to considered as a JEP.
+  They "approve" the submission, assigning the JEP a number, and the submission becomes a "Draft" JEP.
+
+. **<<draft, Draft Status>>** - While the JEP is a "<<draft, Draft>>", Andrea may continue to gather
+  feedback, change the proposal, and record the reasoning and differing views.
+  When she believes the design is complete and represents the consensus of the community,
+  she submits the JEP for review by the <<BDFL>>.
+
+. **<<review, Review>>** - The <<BDFL>> reviews the JEP and decides whether to accept it,
```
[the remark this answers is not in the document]
bitwiseman: @oleg-nenashev @rtyler 
Adding a section to the Reasoning for "Limits of BDFL model".  Resolved?

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: None found but limitation noted.  
RESOLVED

jep/1/README.adoc
bitwiseman: Doesn't have to be.   It means exactly what is says. 😄 
RESOLVED

jep/1/README.adoc
bitwiseman: Out of scope to this JEP. Once we have this JEP active, that might be a new JEP.  😄

jep/1/README.adoc
bitwiseman: Rejected is a pretty strong statement. I'm adding text to this effect.

jep/1/README.adoc
bitwiseman: Updated with additional text. Resolved?

jep/1/README.adoc
bitwiseman: @oleg-nenashev Adding to "Reasoning".  Resolved?

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+[source, asciidoc]
+--
+link:http://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line[One sentence per line]
+--
+
+===== 120 characters per line (REQUIRED)
+
+Aside from the structures mentioned above this point,
+lines MUST be wrapped before 120 characters per line.
+This is required to make diffs and comments reasonably accurate and specific.
```
daniel-beck: This doesn't make any sense. The rendered diff on GH is a useless mess, and the raw diff is even worse.

One line per sentence is clearly the only sensible format. Screen too small? Enable line wrap.

It even assists writing: Line gets too long? Likely a run-on sentence that needs fixing.

daniel-beck:
I wonder whether it would make sense to give certain cross-cutting concerns a defined role in this process.

Today, we have two that would make sense AFAICT:

* Security evaluation
* Infrastructure requirements / side effects

In the future, we would probably also have a UX/usability role. (Someone needs to stop us from adding more system properties whenever we can't reach a consensus.)

I just re-read the current state of this document, and nowhere does it allow e.g. Tyler to veto a proposal because our infra couldn't handle the load, or to tell the sponsor/delegate that the proposal needs reworking because something's fundamentally unsound in terms of security. Sure, we could probably comment on a proposal and try to convince the delegate, but that seems a bit weak.

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+
+|===
+
+
+[[abstract]]
+== Abstract
+
+A Jenkins Enhancement Proposal (JEP) is a design document that
+describes a new feature or aspect of Jenkins itself,
+or the Jenkins Project processes or environment.
```
daniel-beck: Capitalization inconsistent with https://jenkins.io/press/#branding (I prefer this one, but still).

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+describes the motivation for the change,
+and discusses the rationale behind the design.
+
+JEPs are the primary mechanisms for proposing major new
+features, for collecting community input on an issue, and for
+documenting the design decisions that have gone into Jenkins.
+Each JEP must have at least one <<Sponsor>>.
+The JEP Sponsor is responsible for the JEP overall - building
+consensus for that JEP within the community, documenting dissenting opinions,
+coordinating contributors work, and generally ensuring the JEP
```
daniel-beck: missing apostrophe

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+* link:https://github.com/oleg-nenashev[Oleg Nenashev]
+* link:https://github.com/bitwiseman[Liam Newman]
+* link:https://github.com/omehegan[Owen Mehegan]
+* link:https://github.com/orrc[Christopher Orr]
+
+JEP editorship is by invitation of the current editors. All of the JEP workflow
+can be conducted via the GitHub JEP Repository footnoteref:[repo] and pull
+requests.
+
+==== Conributor
```
daniel-beck: Typo, also should be plural for consistency with 'Editors' above?

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+A JEP may have more than one Sponsor, especially after it has been finalized
+and is being maintained over time.
+For simplicity, this document uses the singular
+("JEP Sponsor", "sponsor")
+when referring the one or more people in the role of "Sponsor" of a JEP.
+
+[[requirement-levels]]
+==== Must/Should/May
+
+JEP documents _must_ follow link:https://tools.ietf.org/html/rfc2119[RFC 2119]
```
daniel-beck: 👍

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+==== Overview
+
+Before delving into the details of the JEP workflow,
+let's take a high-level look at how JEP might go.
+
+. **<<start, Inital Discussion>>** - Andrea has an idea for new feature emails jenkinsci-dev@googlegroups.com.
+  She discusses the idea with the group, determining that the idea is worth pursuing.
+  She chooses to be the "<<Sponsor>>" for this potential JEP.
+  She <<discussion, gathers initial feedback>> from the community, adjusts her design as needed,
+  records the reasons for design choices, and keeps track of differing views.
```
daniel-beck: Nowhere is it explicitly stated that someone other than the person proposing something could be its sponsor.

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+and submit pull requests targeting the JEP's feature branch.
+
+[[review]]
+==== JEP Review
+
+Once the sponsor believes a JEP is complete,
+they request the BDFL review the JEP for acceptance, usually via
+an email to the jenkinsci-dev@ mailing list.
+For a JEP that is predetermined to be acceptable
+(e.g., it is an obvious win as-is and/or its implementation has already been checked in)
```
daniel-beck: Huh?

Is that an indirect reference to 'informational' JEPs, or how could this happen?

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+
+
+. **JEP** -- Proposal number, given by the JEP editors. Use "0000" until one is assigned.
+. **Title** -- Brief title explaining the proposal in fewer than 50 characters
+. **Sponsor** -- Sponsor of the JEP, in essence, the individual
+  responsible for seeing the JEP through the process.
+. **Status** -- Draft :speech_balloon:, Deferred :hourglass:, Accepted :ok_hand:,
+  Rejected :no_entry:, Withdrawn :hand:, Final :lock:, Replaced :dagger:, Active :smile:.
+. **Type** -- Describes the type of JEP: Standards, Informational, Process
+. **Created** -- Date (`%Y%m%d`) when the document was first created.
```
daniel-beck: Inconsistent with actual format used above.

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+
+==== Auxiliary Files
+
+JEPs may include auxiliary files such as diagrams. Such files must be
+named appropriately, with lowercase letters and no spaces, and be included in
+the directory with the `README.adoc` describing the JEP.
+
+=== Reporting JEP Bugs, or Submitting JEP Updates
+
+The process for reporting a bug or submiting a JEP update depends on several factors, such
```
daniel-beck: Typo: submiting

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+It occasionally becomes necessary to transfer sponsorship of JEPs to a
+new sponsor. In general, it is preferable to retain the original sponsor as
+a co-sponsor of the transferred JEP, but that's really up to the
+original sponsor. A good reason to transfer sponsorship is because the
+original sponsor no longer has the time or interest in updating it or
+following through with the JEP process, or has fallen off the face of
+the 'net (i.e. is unreachable or not responding to email). A bad
+reason to transfer sponsorship is because the sponsor doesn't agree with the
+direction of the JEP. One aim of the JEP process is to try to build
+consensus around a JEP, but if that's not possible, an sponsor can always
```
daniel-beck: s/an/a

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+=== Transferring JEP Sponsorship
+
+It occasionally becomes necessary to transfer sponsorship of JEPs to a
+new sponsor. In general, it is preferable to retain the original sponsor as
+a co-sponsor of the transferred JEP, but that's really up to the
+original sponsor. A good reason to transfer sponsorship is because the
+original sponsor no longer has the time or interest in updating it or
+following through with the JEP process, or has fallen off the face of
+the 'net (i.e. is unreachable or not responding to email). A bad
+reason to transfer sponsorship is because the sponsor doesn't agree with the
```
daniel-beck: > doesn’t agree with the direction of the JEP

It's unclear to me how this could happen when the sponsor is the author.

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+explain various subsections of Jenkins. JEPs also make it clearer how an
+ambitious new developer to the Jenkins project can propose, and make progress
+upon, a new idea they have for Jenkins.
+
+Overall, less chaos and more productivity is the rationale for JEP.
+
+== Reasoning
+
+The Python community, whose process JEP is modeled after, have successfully
+navigated several large-scale reworkings of Python and it's related tools and
```
daniel-beck: s/it's/its

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+project improvement (technical and otherwise) over time.
+
+The Python process uses "Rationale" as the heading for the section for describing design decisions.
+However the meaning of "Rationale" is similar to "Motivation" in some contexts.
+We decided to use "Reasoning" instead to avoid confusion.
+
+=== Requirements-level terms
+
+Some non-native english speakers commented on the
+<<requirement-levels, Must/Should/May>> mentioned that "should" is a synonym of "must",
```
daniel-beck: Sure, *should* is used the same way your boss *asks* you to *please* do something, but it's understood to not be a request that is easily refused. Just the way some people communicate.

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+The BDFL delegating to other addresses some of that,
+but it is still worth noting possible alternatives.
+
+One alternative would be for the Jenkins Project Governance meeting to have final say.
+Another alternative would be to create some sort of "Technical Steering Commitee",
+separate from Governance, to do this job.
+
+Issues mentioned in relation to these alternatives:
+
+* **Voting policy** - a voting policy woud need to be established,
```
daniel-beck: typo: would

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+=== Limits of BDFL model
+
+People expressed concern over the limits of the current model where the BDFL
+has final say in a number of steps.
+They felt having 1-person bottlenecks in the JEP process could be problematic.
+The BDFL delegating to other addresses some of that,
+but it is still worth noting possible alternatives.
+
+One alternative would be for the Jenkins Project Governance meeting to have final say.
+Another alternative would be to create some sort of "Technical Steering Commitee",
```
daniel-beck: Multiple occurrences of the typo "commitee" in this doc

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+
+Issues mentioned in relation to these alternatives:
+
+* **Voting policy** - a voting policy woud need to be established,
+outlining what percentage of the meeting would need to vote for or against a JEP.
+* **Commitee vs Delegation** - a strictly committee-approval approach
+may not result in good decisions being made in a timely manner. For example,
+only a few people are qualified to make decisions on Remoting.
+It would be difficult for a group of people in `#jenkins-meeting` to vote
+sensibily on a JEP relating to Remoting which most of them don't fully understand.
```
daniel-beck: typo: sensibly

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+Delegation to experts and stakeholders is much more likely to produce high quality improvements.
+* **Lack of established process** - structured technical decision making
+in the Jenkins Project (as outlined in this JEP) is still in its early stages.
+
+The BDFL model with Delegation as needed may not be sufficient in the long run,
+but it will suffice for now.
+
+=== Requiring same license
+
+Some contributors were concerned that changes to a componenent
```
daniel-beck: typo: componenent

jep/1/README.adoc
```diff
@@ -0,0 +1,939 @@
...
+All of those are MIT licensed, such that anybody contributing
+to those repositories is already expected to contribute
+under the existing (MIT) license.
+
+We chose to keep this text as-is until we have a concrete reason to change it.
+
+=== Asciidoc linter
+
+There are currently no Asciidoc linters.
+Should one be found, we will evaluate it for automated enforcemeent of
```
daniel-beck: typo: enforcemeent

rtyler:
> I just re-read the current state of this document, and nowhere does it allow e.g. Tyler to veto a proposal because our infra couldn't handle the load, or to tell the sponsor/delegate that the proposal needs reworking because something's fundamentally unsound in terms of security. Sure, we could probably comment on a proposal and try to convince the delegate, but that seems a bit weak.


For what it's worth, I don't think we need to tweak the language to explicitly say anybody has "veto power" for that. That would naturally happen (and already does) as part of consensus building.

jep/1/README.adoc
bitwiseman: @daniel-beck http://www.dictionary.com/browse/committee

jep/1/README.adoc
bitwiseman: @daniel-beck I'm not sure what action you'd like taken here.

jep/1/README.adoc
bitwiseman: The sponsor doesn't always write the JEP.
One possible case: I come up with an idea, volunteer as sponsor, start writing it, and then three other contributors come along and poke tons of holes in my implementation or some parts of my idea.  I, as sponsor, should not quit being the sponsor because I don't like their implementation.

jep/1/README.adoc
daniel-beck: Unless we're trying to address every review comment in its own section in the doc, I don't think it's useful to have this particular section. The original comment that spawned this directly contradicted dictionaries.

jep/1/README.adoc
bitwiseman: Switching them all to singular. 

RESOLVED.

jep/1/README.adoc
bitwiseman: Yes, or any other reason the BDFL feels they can justify.

bitwiseman:
@daniel-beck  said (in a comment that I can't respond to in the Github UI):

> 
> I wonder whether it would make sense to give certain cross-cutting concerns a defined role in this process.
> 
> Today, we have two that would make sense AFAICT:
> 
> * Security evaluation
> * Infrastructure requirements / side effects
> 
> In the future, we would probably also have a UX/usability role. (Someone needs to stop us from adding more system properties whenever we can't reach a consensus.)
> 
> I just re-read the current state of this document, and nowhere does it allow e.g. Tyler to veto a proposal because our infra couldn't handle the load, or to tell the sponsor/delegate that the proposal needs reworking because something's fundamentally unsound in terms of security. Sure, we could probably comment on a proposal and try to convince the delegate, but that seems a bit weak.

I see your point here.  How strongly do you feel about these?  Security in particular, like backwards compatibility is easy to overlook.   Infra and UX are harder skip because a reference implementation will make them visible.   Would you be willing to create PR with a proposed description of a required "security" section?

daniel-beck:
@bitwiseman It seems to me like you're going more for a 'checklist' approach. I.e. impact on security (or project infrastructure) need to be considered in a JEP, e.g. as new sections in https://github.com/jenkinsci/jep/blob/ebed3f482f5eac8c18c9745d23438ae45c89bbd8/jep/1/README.adoc#required-sections

I think that would be useful anyway.

---

My concern (which is likely hypothetical in any sort of sane consensus building process, but who knows…) that despite concerns of the respective subject matter experts about infra/security impact, something gets accepted, and the negative consequences dumped on those people.

jep/1/README.adoc
bitwiseman: @daniel-beck  

> This doesn't make any sense. The rendered diff on GH is a useless mess, and the raw diff is even worse.

What do you mean here?  The diff for this PR?  Yes, it is insane.  We've made sweeping changes.  

> One line per sentence is clearly the only sensible format. Screen too small? Enable line wrap.

> It even assists writing: Line gets too long? Likely a run-on sentence that needs fixing.

We were avoiding putting any formatting rules in to avoid exactly this situation - nitpicking about which rules are needed or not.  You have a system that you think is best, I have a slightly different one, tyler has another, etc.  

In my experience people with line wrap do not add newlines.  Which then means I as reviewer have to tell them to do so. Also, there are plenty of cases where per sentence breaks result in very long lines and breaking on commas doesn't improve matters (this sentence, for example).  

I chose these rules specifically because they force people to add newline more often.  In my experience, that tends to result in easier diffs.  In the case where the diffs are messy, it indicates lots of change.  

If you'd like to propose some alternative, maybe run some experiments, I'm completely open to changing these.

bitwiseman:
@oleg-nenashev @omehegan @daniel-beck @rtyler  
Let me ask this: what do you think about the idea of moving the discussion of asciidoc formatting to a separate JEP?  The point of this JEP is to establish the JEP process.  Formatting is part of that, but I'm not sure it is required to complete this.  It is important that we address the style rules **soon**, but as a separate JEP, this JEP could refer to those rules and we could apply them to other asciidoc files in the project (such as jenkins.io).

jep/1/README.adoc
bitwiseman: @oleg-nenashev 
Added text regarding Timeliness

bitwiseman:
@daniel-beck 

> @bitwiseman It seems to me like you're going more for a 'checklist' approach. I.e. impact on security (or project infrastructure) need to be considered in a JEP, e.g. as new sections in https://github.com/jenkinsci/jep/blob/ebed3f482f5eac8c18c9745d23438ae45c89bbd8/jep/1/README.adoc#required-sections
> 
> I think that would be useful anyway.

> My concern (which is likely hypothetical in any sort of sane consensus building process, but who knows…) that despite concerns of the respective subject matter experts about infra/security impact, something gets accepted, and the negative consequences dumped on those people.

Exactly.  I'm agreeing with you, we need that section, for the reason you mention and others.
Please submit a PR with the added section description and add it to this JEP as well. I think you're the right person to outline what should be required.

jep/1/README.adoc
oleg-nenashev: RESOLVED

jep/1/README.adoc
oleg-nenashev: RESOLVED

jep/1/README.adoc
oleg-nenashev: RESOLVED

jep/1/README.adoc
bitwiseman: @daniel-beck
Are you saying you object to "this section" as in this part the "Reasoning" isn't needed? 
Or are you saying the "Must/Should/May" section isn't needed?

jep/1/README.adoc
bitwiseman: Updated the Sponsor section to add this. 
RESOLVED.

jep/1/README.adoc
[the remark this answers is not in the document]
bitwiseman: @batmat ^^

bitwiseman:
All the comments on this PR are starting to bog down the UI.  Closing and opening a new PR.

PR #2, opened 2017-09-12
Add an informational JEP draft to capture the Suggested Plugins criteria

rtyler:
Practicing the JEP workflow with @jenkinsci/jep-editors :smile_cat:

### DIFF
--- a/jep/9999/README.adoc
+++ b/jep/9999/README.adoc
@@ -1,0 +1,134 @@
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+= JEP-9999: Criteria for selecting "Suggested Plugins"
+
+:toc:
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 9999
+
+| Title
+| Criteria for selecting "Suggested Plugins"
+
+| Author
+| link:https://github.com/daniel-beck[Daniel Beck], link:https://github.com/rtyler[R Tyler Croy]
+
+| Status
+| :speech_balloon: Draft
+
+| Type
+| Informational
+
+| Created
+| 2017-09-12
+|===
+
+
+[NOTE]
+====
+This proposal/design predates the Jenkins Enhancement Proposal process and has
+therefore been added as an **Information** type of JEP. Normally this would
+have been submitted as a Draft for as a **Standard** type proposal.
+====
+
+== Abstract
+
+Jenkins 2.0 will include a setup dialog that pops up when Jenkins starts for
+the first time. It intends to help people new to Jenkins to get started
+quickly, and present them with a reasonably sized set of plugins to choose from
+for customizing their instance, of they so choose. To this end, not every
+plugin will be presented here. While the extensibility of Jenkins is one of its
+strengths, exposing new users to the 1200+ available plugins before they've
+even started using Jenkins isn't a reasonable approach.
+
+== Specification
+
+Plugins qualify for inclusion in the list based on a number of criteria:
+
+=== Selection Criteria
+
+* Plugin popularity (both in absolute numbers and growth rate), which can serve as a stand-in for plugin usefulness unless we believe this to be inaccurate for specific reasons (e.g. CVS Plugin and Monitor External Job being installed a lot, but only because they are bundled).
+* Does the plugin integrate well with Jenkins? Does it work e.g. with concurrent builds, in secured Jenkins, and distributed Jenkins (master + agent setup)? People new to Jenkins should not be surprised by a plugin offered here undermining or not working with these features.
+* Is the plugin more than just superficial themeing/UI? While those have their place and are popular, advertising plugins changing the UI out of the box would be very weird (Simple Theme Plugin, Green Balls)
+* Are the features it provides useful to someone fairly new to Jenkins? If the plugin requires someone to e.g. be familiar with Jenkins internals before they can use it successfully, or to have multiple Jenkins instances, it may not be a good choice for this dialog. For example, if the script console were a plugin, it would probably not be on here.
+* Even if it's not popular, is it a reasonably common SCM? SCMs are so fundamental to successful Jenkins use that these shouldn't be as heavily restricted as some of the other categories.
+* The plugin most work without requiring a restart of Jenkins.
+* The plugin must be discoverable when installed, and at least allow users to get started without outside documentation. (This disqualifies Build Token Root Plugin.)
+* The plugin is hosted in the link:https://github.com/jenkinsci[jenkinsci] GitHub organization
+* The plugin must be maintained reasonably well (e.g. was updated in the last 18 months or so, and ideally has a dedicated maintainer)
+
+=== Categories
+
+Given the diverse feature set of Jenkins plugins, it is difficult to define a
+set of categories that can be used to group plugins without misrepresenting
+them (and having every plugin only how up once in the list, rather than
+multiple times). The following categories appear to adequately provide a
+high-level view of plugin functionality in Jenkins:
+
+* **General**: Anything that has no better category. Depending on contents, these may result in the creation of new categories. Possible examples are Rebuilder Plugin, Timestamper Plugin, or Translation Assistance Plugin.
+* **Organization and Administration**: Plugins that help Jenkins administrators accomplish their goals, and that help keep order. Examples are Monitoring Plugin, Antisamy Markup Formatter, or Disk Usage Plugin.
+* **Build Features** are plugins that provide features to builds/jobs typically beyond an SCM, build step or publisher. SSH Agent Plugin, AnsiColor Plugin, and Timestamper.
+* **Build Tools**: for plugins providing build tool integration, e.g. NodeJS Plugin, Maven Plugin, Gradle Plugin.
+* **Build Analysis and Reporting**: Plugins in this category process reports or data generated during builds, such as JUnit Plugin, Build Failure Analyzer, or Checkstyle Plugin.
+* **Build Pipelines and Continuous Delivery**: is for plugins that allow creating more complex chains of actions that would reasonably be possible in a single freestyle job. Examples are Copy Artifact Plugin, Parameterized Trigger Plugin, or Workflow Plugin.
+* **SCM**:provides a selection of plugins that integrate with SCM tools, e.g. Subversion Plugin, GitHub Plugin, or GitHub Pull Request Builder Plugin.
+* **Distributed Builds and Containers**: are plugins that provide different build environments to builds, such as Docker Plugin, Windows Slaves Plugin, or Matrix Project Plugin.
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Do you propose to drive JEPs over pull requests? It is definitely better than email tracks, but I would argue that we could benefit from review features of tools like Google Doc (especially "propose changes" mode).

rtyler:
@oleg-nenashev please consult #1 for the JEP editor's responsibilities and workflow.

rtyler:
@oleg-nenashev I suppose put less bluntly, this is a practice run. So I would expect us first to make sure that the document is correct/understandable (not judging the JEP itself).

Then create a number for it, push this into a branch, and then the real discussion can happen.

The way I imagine this happening on a `jep-2` branch would be contributors would be able to propose changes to the document via PRs to `jep-2` rather than simply comments on the PR.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+strengths, exposing new users to the 1200+ available plugins before they've
+even started using Jenkins isn't a reasonable approach.
+
+== Specification
+
+Plugins qualify for inclusion in the list based on a number of criteria:
+
+=== Selection Criteria
+
+* Plugin popularity (both in absolute numbers and growth rate), which can serve as a stand-in for plugin usefulness unless we believe this to be inaccurate for specific reasons (e.g. CVS Plugin and Monitor External Job being installed a lot, but only because they are bundled).
```
jglick: *were* bundled

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* Does the plugin integrate well with Jenkins? Does it work e.g. with concurrent builds, in secured Jenkins, and distributed Jenkins (master + agent setup)? People new to Jenkins should not be surprised by a plugin offered here undermining or not working with these features.
+* Is the plugin more than just superficial themeing/UI? While those have their place and are popular, advertising plugins changing the UI out of the box would be very weird (Simple Theme Plugin, Green Balls)
```
jglick: Green Balls actually includes colorblindness support.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* Are the features it provides useful to someone fairly new to Jenkins? If the plugin requires someone to e.g. be familiar with Jenkins internals before they can use it successfully, or to have multiple Jenkins instances, it may not be a good choice for this dialog. For example, if the script console were a plugin, it would probably not be on here.
+* Even if it's not popular, is it a reasonably common SCM? SCMs are so fundamental to successful Jenkins use that these shouldn't be as heavily restricted as some of the other categories.
+* The plugin most work without requiring a restart of Jenkins.
```
jglick: Not possible in all cases. Let us say that if it requires a restart, there is a solid technical reason for it.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* The plugin must be discoverable when installed, and at least allow users to get started without outside documentation. (This disqualifies Build Token Root Plugin.)
```
jglick: Unless that were to add `_api.jelly` pages documenting its usage, I suppose.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* The plugin is hosted in the link:https://github.com/jenkinsci[jenkinsci] GitHub organization
```
jglick: missing `.`

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* The plugin must be maintained reasonably well (e.g. was updated in the last 18 months or so, and ideally has a dedicated maintainer)
```
jglick: ditto

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+
+=== Categories
+
+Given the diverse feature set of Jenkins plugins, it is difficult to define a
+set of categories that can be used to group plugins without misrepresenting
+them (and having every plugin only how up once in the list, rather than
```
jglick: _show_ up

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+multiple times). The following categories appear to adequately provide a
+high-level view of plugin functionality in Jenkins:
+
+* **General**: Anything that has no better category. Depending on contents, these may result in the creation of new categories. Possible examples are Rebuilder Plugin, Timestamper Plugin, or Translation Assistance Plugin.
+* **Organization and Administration**: Plugins that help Jenkins administrators accomplish their goals, and that help keep order. Examples are Monitoring Plugin, Antisamy Markup Formatter, or Disk Usage Plugin.
+* **Build Features** are plugins that provide features to builds/jobs typically beyond an SCM, build step or publisher. SSH Agent Plugin, AnsiColor Plugin, and Timestamper.
```
jglick: You just mentioned Timestamper as an example of something with no category.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* **Build Tools**: for plugins providing build tool integration, e.g. NodeJS Plugin, Maven Plugin, Gradle Plugin.
+* **Build Analysis and Reporting**: Plugins in this category process reports or data generated during builds, such as JUnit Plugin, Build Failure Analyzer, or Checkstyle Plugin.
+* **Build Pipelines and Continuous Delivery**: is for plugins that allow creating more complex chains of actions that would reasonably be possible in a single freestyle job. Examples are Copy Artifact Plugin, Parameterized Trigger Plugin, or Workflow Plugin.
```
jglick: Pipeline!

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* **SCM**:provides a selection of plugins that integrate with SCM tools, e.g. Subversion Plugin, GitHub Plugin, or GitHub Pull Request Builder Plugin.
```
jglick: `github` is a grab bag of features actually, not an SCM plugin per se. `github-branch-source` might be a better example.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+* **Distributed Builds and Containers**: are plugins that provide different build environments to builds, such as Docker Plugin, Windows Slaves Plugin, or Matrix Project Plugin.
+** Currently no container plugins made the cut.
```
jglick: What would some examples be? VMWare?

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+includes the ability to choose plugins, it's better not to overwhelm
+inexperienced users by exposing them to the full list of plugins from the
+start.
+
+Forcing those with no experience to choose between 1000+ plugins before they
+can even start using Jenkins is an impossible task. They may not even be fully
+aware what plugins are for, and that Jenkins doesn't really do much without any
+of them.
+
+Who points them to Folders, Timestamper, or Build Timeout Plugin? Sure, they
```
jglick: Not sure what point is being made here.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+can always accept the defaults, but what if they want to modify something that
+should be simple enough, like add support for a less popular SCM like
+Mercurial?
+
+A selection of known good plugins is a much better choice for this particular
+setup dialog, and that's how it's currently implemented. The initial selection
+of 'curated' (= available for selection) plugins was provided to Gus and Tom by
+me (and tweaked by them after some early reviewer feedback). I used the stats
```
jglick: By whom?

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+of them.
+
+Who points them to Folders, Timestamper, or Build Timeout Plugin? Sure, they
+can always accept the defaults, but what if they want to modify something that
+should be simple enough, like add support for a less popular SCM like
+Mercurial?
+
+A selection of known good plugins is a much better choice for this particular
+setup dialog, and that's how it's currently implemented. The initial selection
+of 'curated' (= available for selection) plugins was provided to Gus and Tom by
```
jglick: To whom?

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+therefore been added as an **Information** type of JEP. Normally this would
+have been submitted as a Draft for as a **Standard** type proposal.
+====
+
+== Abstract
+
+Jenkins 2.0 will include a setup dialog that pops up when Jenkins starts for
+the first time. It intends to help people new to Jenkins to get started
+quickly, and present them with a reasonably sized set of plugins to choose from
+for customizing their instance, of they so choose. To this end, not every
```
omehegan: "if they so choose"

rtyler:
@jglick thanks for the review. Honestly I pretty swiftly copied and pasted the original wiki page which @daniel-beck wrote.

I'll address these after we give it a number,  squash it into a `jep-X` branch.

jep/9999/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+even started using Jenkins isn't a reasonable approach.
+
+== Specification
+
+Plugins qualify for inclusion in the list based on a number of criteria:
+
+=== Selection Criteria
+
+* Plugin popularity (both in absolute numbers and growth rate), which can serve as a stand-in for plugin usefulness unless we believe this to be inaccurate for specific reasons (e.g. CVS Plugin and Monitor External Job being installed a lot, but only because they are bundled).
+* Does the plugin integrate well with Jenkins? Does it work e.g. with concurrent builds, in secured Jenkins, and distributed Jenkins (master + agent setup)? People new to Jenkins should not be surprised by a plugin offered here undermining or not working with these features.
```
omehegan: As I read this, it sounds like people shouldn't act surprised if a plugin causes problems. Maybe say "should not experience a plugin here undermining..."

jep/9999/README.adoc
omehegan: "theming"

jep/9999/README.adoc
omehegan: "must work"

jep/9999/README.adoc
omehegan: Space after colon.

jep/9999/README.adoc
daniel-beck: Which is only necessary because it introduces the red/green color blindness problem. I don't remember complaints about blue/yellow or other issues in the default setup (perhaps because of brightness differences).

bitwiseman:
JEP-2 approved as "Draft".
https://github.com/jenkinsci/jep/tree/jep-2
@rtyler  will update with review comments.

jep/9999/README.adoc
rtyler: This JEP is only informational with the process which was previously agreed to.

I think if we wish to change this, it would be best to include this consideration in a newer proposal which revisits the Suggested Plugin criteria entirely (which, honestly, I think @daniel-beck or I might end up writing in the very near future for other reasons :smile_cat: )

PR #8, opened 2017-09-19
Add some more details based on @oleg-nenashev's initial review

rtyler:


jep/1/README.adoc
```diff
@@ -398,7 +409,9 @@ For a JEP to be "Accepted" it must meet certain minimum criteria:
 
 * It must be a clear and complete description of the proposed enhancement.
 * The enhancement must represent a net improvement.
-* The proposed implementation, if applicable, must be solid and must not complicate Jenkins unduly.
+* The proposed implementation, if applicable, must be solid, must not
+  complicate Jenkins unduly, and must be the same license as the component the
+  proposal is meant to added to (or MIT licensed by default).
```
oleg-nenashev: "same license" is maybe too strong. But it's better than legal holywars about "compatible licenses".

OTOH some exceptions could be useful. E.g. some companies strictly require "Apache v.2", because MIT is not explicit about the patent release. By setting this condition we explicitly require companies to do reference implementation with the MIT license, and hence such potential contributors may have no way to comply with the process without getting a legal-approved exception (it hurts :( ).

I would just per-approve MIT, Apache and BSD licenses for reference implementations, independently of the component license. GPL/LGPL is more complicated in the case of Jenkins.

PR #9, opened 2017-09-22
Address more of feedbackJep 1

bitwiseman:
Address more of the feedback for @oleg-nenashev.

### DIFF
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -19,7 +19,7 @@ endif::[]
 | Title
 | Jenkins Enhancement Proposal Format
 
-| Author
+| Sponsor
 | link:https://github.com/rtyler[R Tyler Croy]
 
 | Status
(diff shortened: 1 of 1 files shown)


jep/1/README.adoc
```diff
@@ -758,6 +851,10 @@ The Python process uses "Rationale" as the heading for the section for describin
 However the meaning of "Rationale" is similar to "Motivation" in some contexts.
 We decided to use "Reasoning" instead to avoid confusion.
 
+Some non-native english speakers commented on the
```
oleg-nenashev: Not sure it is really required to keep it in the body of the document.
Maybe we could create some Wiki pages for extra clarification of the top of the JEP document (and extra examples).

jep/1/README.adoc
```diff
@@ -19,7 +19,7 @@ endif::[]
 | Title
 | Jenkins Enhancement Proposal Format
 
-| Author
+| Sponsor
```
oleg-nenashev: "Sponsor" is better than "Author" for sure. It may be also a bit ambiguous because it's also may be considered as a "person who pays the bill". OTOH other projects use "sponsor" for similar purposes, so I am fine with that.

bitwiseman: @oleg-nenashev - I agree, it is imperfect.  But it avoids "Owner" and "Author".  It is sponsor more in the sense of "Someone who vouches for and guides".   

RESOLVED.

jep/1/README.adoc
bitwiseman: @oleg-nenashev
I lean towards  including at least a highlight of notable comments about a choice. I feel this kind of comment is exactly what this section is for - part of the "documenting differing opinions".  

Other non-native english speakers will appreciate that we noted this issue.  If someone comes along later with a strong  opinion about this, perhaps considering looking for better options, they will be able to include this in their considerations.

Also, it is a small enough comment that pushing it out of the document body would mean it would almost surely be overlooked.

PR #17, opened 2017-10-17
Added Reviewer role and updated BDFL

bitwiseman:
@jenkinsci/jep-editors 

Having let some previous feedback percolate for a bit, I thought it would be better to have a role that was "JEP Reviewer" instead of "BDFL-delegate".   I also expanded the text for the BDFL role to make it clearer that they have great power and a responsibility to generally avoid using it.

### DIFF
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -115,7 +115,21 @@ There are three kinds of JEP:
 ==== BDFL
 
 There are several references in this JEP to the "BDFL". This acronym stands for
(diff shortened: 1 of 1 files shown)


rtyler:
> for making the text explicit. I am going to vote against BDFL concept with such wording, but it makes sense to merge it into the JEP to make the explicit and discussable

@oleg-nenashev can you please provide a proposal for what you would be comfortable with. I understand your concerns with a BDFL codification, but I haven't yet seen a counter-proposal formed from anybody.

PR #13, opened 2017-09-26
Prefer PRs to comments

bitwiseman:
This is an example of the idea I'd like to try.  As noted in #12, and in the comments below I'd like to recommend (or at least describe) in JEP-1 how changes to a JEP can be suggested, refined, and merged to a JEP Draft. 

The intent of this idea was motivated by the limitations of the Github PR UI, but I'm wondering if it might have other benefits.  
* Quicker turn-around - If people spot more than a few typos, submitting a PR to fix all of them is about the same amount of effort for the reviewer as making the comments and is much easier for the sponsor or other contributors. 
* Focus on outcomes - Comments seem guide people toward asking if something can be done rather than doing it.  As with typos, talking about doing something often takes more effort than writing something concrete, submitting it as a PR, and doing a couple iterations on that PR. 
* Alternate views easy to add to "Reasoning" - I ended up adding several topic to the Reasoning section based on comments.  If these comments had been in their own PR, it would have been easier for those who proposed the ideas to move them to the Reasoning section for merging, rather than bottlenecking at one contributor making changes.

This is obviously a work in progress.  I'd like feedback before trying to hammer out details.

@jenkinsci/jep-editors @daniel-beck

PR #18, opened 2017-10-18
Removed BDFL

bitwiseman:
@oleg-nenashev @rtyler 
After some review and discussion, I've removed the BDFL from the JEP.

### DIFF
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -112,27 +112,6 @@ There are three kinds of JEP:
 
 === JEP Terminology
 
-==== BDFL
-
-There are several references in this JEP to the "BDFL". This acronym stands for
-"link:https://en.wikipedia.org/wiki/Benevolent_dictator_for_life[Benevolent Dictator for Life]".
-The BDFL is a single person whose role in the project is to act as decision maker
-when one is needed. The responsibilities of the BDFL are:
-
-* Review JEPs and decide whether they will will be accepted
-* Delegate JEP Reviewer responsibilities to other contributors as appropriate
-* Resolve disputes or arguments that cannot be resolved by consensus
-* Take any other actions as part of the JEP process that they deem best of the community
-* Clearly communicate the reasoning for any actions taken or decisions made
-* Refrain from using their power, letting the community self-govern whenever possible
-
-The BDFL may override the decision any other contibutors.
-Decisions made by the BDFL are final and the community agrees to abide by them.
-
-For the Jenkins project the BDFL is
-link:https://github.com/kohsuke[Kohsuke Kawaguchi],
-original creator of Jenkins.
-
 ==== Editor
 
 The JEP editors are individuals responsible for managing the administrative
@@ -153,7 +132,7 @@ requests.
 ==== Contributor
 
 A JEP may have any number of "Contributors" who
-help write, implement, discuss, or offer feedback about that JEP.
+help write, implement, discuss, review, or offer feedback about that JEP.
 One contributor might do only one or any combination of these
 of these things during any part of the life of a JEP.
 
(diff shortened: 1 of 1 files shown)


jep/1/README.adoc
```diff
@@ -316,15 +298,15 @@ the idea is applicable to the entire community and not just the sponsor. Just
 because an idea sounds good to the sponsor does not mean it will work for most
 people in most areas where Jenkins is used.
 
+A <<JEP Reviewer>> should be selected before the JEP is submitted,
+but may also be added or changed any time before a document is reviewed.
```
oleg-nenashev: I see no specific need in this requirement, but OK.
I would just say that "JEP Reviewers should be selected via a discussion in the mailing list or in the Governance meeting, discussion in the pull request is NOT enough"

jep/1/README.adoc
```diff
@@ -546,6 +523,21 @@ If one or more contributors are not responding
 and the sponsor chooses to move forard without their feedback,
 they should document that choice in the "<<Reasoning>>" section of the JEP.
 
+==== Resolving Disputes
```
oleg-nenashev: 👍

jep/1/README.adoc
```diff
@@ -546,6 +523,21 @@ If one or more contributors are not responding
...
+
+Except for the decision by a reviewer to accept or reject a JEP,
+the JEP process is run by
+link:https://en.wikipedia.org/wiki/Consensus_decision-making[consensus].
+It is the responsibility of the every contributor to respect other contributors,
+listen to their perspectives, and attempt to find solution that work for everyone.
```
oleg-nenashev: "a solution that works" or "solutions that work"

jep/1/README.adoc
```diff
@@ -546,6 +523,21 @@ If one or more contributors are not responding
...
+If consensus cannot be achieved or if there is strong disagreement with the decision of a reviewer,
+a contributor may request that the Jenkins Governance Board intervene.
+The board will take up the matter at their next regular meeting and render a decision at the meeting following.
```
oleg-nenashev: Just to clarify, do you mean "Jenkins Governance Board" or "Jenkins Governance Meeting"? These are different entities. Regular && public meetings apply only to the latter one.

jep/1/README.adoc
```diff
@@ -546,6 +523,21 @@ If one or more contributors are not responding
...
+
+This should be a last resort and should happen extremely rarely.
+
```
oleg-nenashev: "Governance meeting may make a decision to return JEP to draft and to assign another reviewer".
I am not sure whether Governance Meeting should be a place for "accepting"/"rejecting" JEPs though formally it should be eligible to do that as an ultimate power entity in Jenkins org.

jep/1/README.adoc
```diff
@@ -853,39 +848,13 @@ The Python process uses "Rationale" as the heading for the section for describin
...
-
-The BDFL model with Delegation as needed may not be sufficient in the long run,
-but it will suffice for now.
+The Python process has a role called the BDFL,
+the "link:https://en.wikipedia.org/wiki/Benevolent_dictator_for_life[Benevolent Dictator for Life]".
+The BDFL acts as reviewer and resolver of disuputes.
+People expressed concern over the limits of the BDFL-based model
+They felt having 1-person bottlenecks or too much power in the JEP process could be problematic.
+We chose to remove the BDFL role, adding a "Reviewer" role to handle JEP review and resolution,
+and directing any disputes to the Jenkins Governance Board.
```
oleg-nenashev: s/Board/Meeting?

bitwiseman:
@oleg-nenashev Great, thanks!  Instead of asking what I meant or if something should be changed, it would be super helpful if you could submit the changes you'd believe need to happen to this PR as a PR of your own that targets this PR.   see https://github.com/jenkinsci/jep/pull/13.  

For example, Board vs Meeting.  I don't know.  What do you think?  Which ever answer it is, the PR would be more effective.  You can either change it or add clarification as to why it is this one.

jep/1/README.adoc
bitwiseman: I considered adding some statement that said JEPs could be rejected or returned to draft via this path, but not accepted.  Only reviewers can accept, but as I said below, please go write it up as you believe it should be.

oleg-nenashev:
> For example, Board vs Meeting. I don't know. What do you think? Which ever answer it is, the PR would be more effective. You can either change it or add clarification as to why it is this one.

I would rather prefer "Governance Meeting", because it's more public. Also there is a reelection issue in the board, which I would prefer to avoid in the JEP process. 

OTOH Board resolves conflicts between contributors in Code-of-Conduct and hence resolving JEP conflicts there would be natural.

I will create a pull request

bitwiseman:
Due to objections by @kohsuke, I'm closing this.  See #19.

PR #12, opened 2017-09-26
DRAFT: Jenkins Enhancement Proposal

bitwiseman:
#1 was beginning to bog down the browser UI due to all the comments.

PR #21, opened 2017-10-30
JIRA

jglick:
Proposed addition to #12.


jglick:
I did not wish to make a mandatory change to the process. This at least offers a standard field for JEP sponsors who wish to opt in to JIRA association.

Anyway informational JEPs probably do not benefit from linking to JIRA, etc.

PR #15, opened 2017-09-26
Added Security and Infra

bitwiseman:
Per discussion on IRC

PR #24, opened 2017-11-01
Add testing section

jglick:
Initially proposed by @oleg-nenashev in https://github.com/jenkinsci/jep/pull/15#pullrequestreview-65480476 but I independently noticed that this was absent and generally desirable.

See [this section](https://gist.github.com/jglick/9721427da892a9b2f75dc5bc09f8e6b3#testing) of a pre-JEP proposal of mine for an example.

### DIFF
--- a/jep-template/0000/README.adoc
+++ b/jep-template/0000/README.adoc
@@ -153,6 +153,27 @@ If this proposal will have no impact on infrastructure, this section may simply 
 There are no new infrastructure requirements related to this proposal.
 ====
 
+== Testing
+
+[TIP]
+====
+If the JEP involves any kind of behavioral change to code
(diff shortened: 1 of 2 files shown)


jep-template/0000/README.adoc
```diff
@@ -153,6 +153,27 @@ If this proposal will have no impact on infrastructure, this section may simply
 There are no new infrastructure requirements related to this proposal.
 ====
 
+== Testing
+
+[TIP]
+====
+If the JEP involves any kind of behavioral change to code
+(whether in a Jenkins product or backend infrastructure),
```
oleg-nenashev: "Everything exception Informational JEP"? The current wording LGTM though

recena:
@jglick It makes sense.

PR #25, opened 2017-11-01
Suggesting that the merge to master occur when a JEP is finalized

jglick:
#12 apparently never said if or when a `jep-nnnn` feature branch would be merged into `master`. I am picking finalization time as a first pass, though it might make more sense to put it in `master` as soon as it is accepted.

(Frankly I was a little surprising to find that the merge to `master` does not happen as soon as the _draft_ is accepted. That is after all the point at which you are committed to a particular number, so as a record of this JEP’s existence and status IMO it should be there already, even if it was ultimately rejected or abandoned.)

### DIFF
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -485,6 +485,8 @@ When the implementation is complete and incorporated into the
 appropriate "main" code repository, the JEP sponsor will change
 the JEP's status changed to "Final".
 
+At this time the JEP editors will merge its feature branch into the `master` branch of the `jep` repository.
+
 [[maintenance]]
 ==== JEP Maintenance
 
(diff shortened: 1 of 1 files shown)


bitwiseman:
@jglick  
You are correct about wanting JEPs to be recorded even if rejected.  But I also think we don't want to merge to master while in draft, since as you've seen, they may got through _a lot_ of changes.  Having all that iterating off to the side seems better to me.  

Maybe take a few minutes and propose what you think it really should look like and then other can counter-propose/amend?

jglick:
> I also think we don't want to merge to master while in draft, since as you've seen, they may got through _a lot_ of changes.

And…so? Is there some problem with having lots of PRs filed against the `master` branch? Who cares?

jglick:
> propose what you think it really should look like

#29

bitwiseman:
@jglick  Closing this in favor of #29.

PR #26, opened 2017-11-01
Fixing number

jglick:
I guess @bitwiseman forgot this.


bitwiseman:
@jglick - I'm going to have to write a script that does the right thing. 
Thanks.

PR #29, opened 2017-11-06
Simplified branch handling

jglick:
As compared to #25, this is what I actually want, not what I tried to infer someone else wanted.

### DIFF
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -394,7 +394,7 @@ the editors will approve the submission as a draft JEP
 by following the steps outlined in the
 <<approve-as-draft, editors' "Approve as Draft" section>>.
 When they are done, the Draft JEP will have an official JEP number and
-the sumbmission PR will have been merged to a matching folder
+the submission PR will have been merged to a matching folder
 (for example,
 `link:https://github.com/jenkinsci/jep/tree/master/jep/1[jep/1]`)
 in the `master` branch.
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -394,10 +394,10 @@ the editors will approve the submission as a draft JEP
 by following the steps outlined in the
 <<approve-as-draft, editors' "Approve as Draft" section>>.
 When they are done, the Draft JEP will have an official JEP number and
-the sumbmission PR will have been merged to a matching folder and feature branch
-(i.e the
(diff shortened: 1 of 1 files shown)


rtyler:
My original motivation for not landing things left and right in the `master` branch is that the `master` branch in the current JEP-1 indicates more or less what's been approved without requiring much searching around.

Landing documents in various stages of the workflow doesn't bother me, but this to me is strictly an editor workflow question and doesn't really tangibly affect most people.

Therefore I would like a judgement call from the other @jenkinsci/jep-editors

daniel-beck:
> doesn't really tangibly affect most people

What's the expected workflow of an active project contributor wanting to review and provide feedback to in-progress JEPs? Would every such JEP always have a corresponding PR? Going through branches seems cumbersome.

rtyler:
> What's the expected workflow of an active project contributor wanting to review and provide feedback to in-progress JEPs? Would every such JEP always have a corresponding PR? Going through branches seems cumbersome.

Based on how we have handled JEP-1, there would be an open PR to master branch for each branch in my mind which provides a central review point

bitwiseman:
@rtyler - But as we saw with JEP-1, that means that PR gets really long and cumbersome.  It also encourages people to comment on the PR instead of submitting PR's and discussing on the dev-list. 

In this "master branch", the central point of review for all Draft JEPs is the `HEAD` of `master` and then consider any open PRs.

jglick:
> script(s) could automatically generate an overview file (or update the readme) indicating the current status of all JEPs

Yes! To me it just seems much simpler to be able to tell people: look in this repo (GitHub will select `tree/master/` by default) and you will see all JEPs, listed by number. As soon as a JEP is accepted as a draft and gets a number it appears here and never goes away. If we want to develop some tool to produce a summary by status [like the JCP does](https://jcp.org/en/jsr/stage), great.

(At some point we may wish to add a metadata field indicating a target release date, where applicable, which would allow us to produce something like an actual roadmap for Jenkins as a whole for the first time. Currently I think only [Blue Ocean has any equivalent](https://jenkins.io/projects/blueocean/roadmap/).)

omehegan:
I reviewed this at @rtyler's request and I think it sounds reasonable. I don't have a strong opinion.

bitwiseman:
Editors seem to be agreed.  `master` is good.

PR #39, opened 2017-12-05
s/ewelinawilkosz2/ewelinawilkosz/

jglick:
@ewelinawilkosz2 does not seem to exist while @ewelinawilkosz does and claims to be from @Praqma, so…? CC @ndeloof


ewelinawilkosz:
Yes, sorry, I made a mess with accounts long time ago and decided to fix it before I start serious contributing... I explained that to @ndeloof on slack. Hope it won’t cause more issues

PR #40, opened 2017-12-15
Accepting JEP #201 as BDFL

kohsuke:
This one IMO is no brainer because ...

* People who worked in this area before, including myself, are very happy to see this effort
* From what I've heard from @ewelinawilkosz and @ndeloof, people who have seen it are excited as well, not to mention the popularity of [JENKINS-31094](https://issues.jenkins-ci.org/browse/JENKINS-31094)
* [To the extent this was discussed in the dev list](https://groups.google.com/d/msg/jenkinsci-dev/_l8UBjCP1vQ/65nC-B7LBgAJ), reaction has been very supportive
* Effort is well underway

### DIFF
--- a/jep/201/README.adoc
+++ b/jep/201/README.adoc
@@ -24,9 +24,9 @@ endif::[]
 | Status
 // Uncomment the appropriate line.
 //| Not Submitted :information_source:
-| Draft :speech_balloon:
+//| Draft :speech_balloon:
 //| Deferred :hourglass:
-//| Accepted :ok_hand:
+| Accepted :ok_hand:
 //| Rejected :no_entry:
 //| Withdrawn :hand:
 //| Final :lock:
@@ -232,3 +232,4 @@ https://github.com/jenkinsci/configuration-as-code-plugin
 == References
 
 This topic was initially discussed on https://issues.jenkins-ci.org/browse/JENKINS-31094[JENKINS-31094].
+Subsequent discussion https://groups.google.com/d/msg/jenkinsci-dev/6TjlxEqHUEs/nKSG1xSkCQAJ[here].


jep/201/README.adoc
```diff
@@ -232,3 +232,4 @@ https://github.com/jenkinsci/configuration-as-code-plugin
 == References
 
 This topic was initially discussed on https://issues.jenkins-ci.org/browse/JENKINS-31094[JENKINS-31094].
+Subsequent discussion [here](https://groups.google.com/d/msgid/jenkinsci-dev/CANfRfr2BpTRM40fMWHG_7RhQfWdNocomzmkR7%3DdR0xO8SqkUYw%40mail.gmail.com).
```
daniel-beck: This is AsciiDoc, not Markdown.

Also, the canonical URL to the thread is https://groups.google.com/d/msg/jenkinsci-dev/6TjlxEqHUEs/nKSG1xSkCQAJ

daniel-beck:
Acceptance criteria: https://github.com/jenkinsci/jep/tree/master/jep/1#accepting-a-jep

Given this PR, I'm assuming the draft JEP in its current form is not a work in progress (AFAICT there's no way for me to determine that distinction). So I re-read it and have a few concerns I hope @ndeloof could still address. It's not clear to me where those should be published, so I'm posting them here.

* On one hand, this effort requires data binding to be used everywhere, on the other it does not propose to deprecate the `newInstance` approach to descriptor implementation. There's a conflict between the reality of that approach being common, and the "no-glue-code requirement" that's not being resolved.
* There's no data supporting going with a data binding-only approach can even achieve the stated goal to a reasonable degree: For example, that a large majority of plugins already use data binding, or that it's trivial to replace all such implementations with a data-bound one. I would expect an overview of plugin changes needed to be specifically called out for an effort like this, if applicable.
* The proposal reasons that the configuration as code effort should mimic the web UI as closely as possible, when a major subproject is basically working to retire the existing web UI. So, why even do this?

Nice to have:

* Internal code terms may be different from the UI labels (think agent/slave). Would it be desirable to specifically have alias support for inconsistencies like this? This doesn't even take into account i18n related challenges.

I think it would be useful to remind everyone that this draft is considered finished by its authors, and should be re-reviewed with an expectation that the current state will be the final one.

ndeloof:
> This effort requires data binding to be used everywhere, on the other it does not propose to
> deprecate the newInstance approach to descriptor implementation. There's a conflict between the 
> reality of that approach being common, and the "no-glue-code requirement" that's not being
> resolved.

The current prototype relies on data-binding. `newInstance` is not impacted as long as a plugin doesn't override the default implementation for custom binding. So for Describables we have no issue in most cases. Descriptor#configure on the other side (i.e global.jelly we want to instrument)
don't use to rely on databinding. 

> There's no data supporting going with a data binding-only approach can even achieve the stated
> goal to a reasonable degree: For example, that a large majority of plugins already use data 
> binding, or that it's trivial to replace all such implementations with a data-bound one. I would 
> expect an overview of plugin changes needed to be specifically called out for an effort like this, 
> if applicable.

That's not the case. We noticed many plugins do already have setters for internal global configuration model (most probably for groovy script support), so we rely on those and can support plugins out-of-the-box, but for few configuration option. So the "_no glue code_" promise is not yet 100% reached, just need to get plugins to follow some best-practices.
We published [this guide](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/PLUGINS.md) and [sample pull request](https://github.com/jenkinsci/mailer-plugin/pull/39). 


> The proposal reasons that the configuration as code effort should mimic the web UI as closely as 
> possible, when a major subproject is basically working to retire the existing web UI. So, why even
> do this?

Because user do know the web UI. Which subproject are you referring to ? is it BlueOcean ?
I don't think BO do cover jenkins global configuration, only jobs and only for pipeline definition (so far)


> Nice to have:
> Internal code terms may be different from the UI labels (think agent/slave). Would it be desirable
> to specifically have alias support for inconsistencies like this? This doesn't even take into account
> i18n related challenges.

I consider this a _Must Have_  as some attribute don't just use politically debated names, but internal type names, like `labelString` that makes no sense for end-user trying to configure a `label`.
see https://github.com/jenkinsci/structs-plugin/pull/18.

daniel-beck:
@ndeloof Thanks for your response and references. I think some of that would be nice to address in the JEP for future readers.

ndeloof:
@daniel-beck I could prepare a PR to amend the JEP doc. But I wonder if such "implementation details" better fit in the configuration-as-code repository. Shouldn't the JEP focus on goals / scope ?

daniel-beck:
@ndeloof Unsure, perhaps a JEP editor can help. The only other technical JEP we have so far is https://github.com/jenkinsci/jep/tree/master/jep/200 and that goes into much more detail.

Note that my questions are more about the specification. Concerns about how to achieve coverage, whether to deprecate non-databound objects, how to deal with inconsistency between labels and code, and tying the configuration format to what amounts to a legacy UI are not implementation details to me.

ndeloof:
@daniel-beck proposed doc update here : https://github.com/jenkinsci/jep/pull/41

jep/201/README.adoc
kohsuke: Fixed

kohsuke:
For an informational JEP that curves out a space for a sub-project, I think the current level of description is preferable. It should cover goals and scopes that help unite potential contributors, because committing to the goal is more fundamental than committing to a specific implementation approach.

That goal statement then creates technical conversations like this, which is great, so that we not only have a viable path forward but that's visible for others.

bitwiseman:
@daniel-beck - Have all your concerns have been sufficiently addressed?  Do you agree that the prototype implementation is sufficiently complete to accept this JEP?

daniel-beck:
@bitwiseman None of my remarks are blocking this in any way. Note that I kept the 'request changes' review about the broken formatting, and my comments regarding proposal contents, clearly separated.

kohsuke:
I thought more about this.

I think the point being raised isn't exactly whether the state of the effort constitutes "sufficiently complete prototype." That criteria makes a clearer sense in the context of a core change, but I think this case is bit unique because it's somewhere between informational JEP and standard JEP.

It's somewhat like informational JEP because it captures the goals for a sub-project and therefore by itself that isn't really binding. But the purpose of capturing that as a JEP is in part because the effort aims to cover sufficiently large and "right" use cases to aim for THE CaC solution for Jenkins, and that aspect resonates with this being a standard track JEP.

The heart of the issue that @daniel-beck raises is that the rest of Jenkins developers need to see the technical assumptions and approaches baked into this project more clearly in order to provide meaningful feedback, the kind that this JEP is seeking to solicit in order to aim for THE CaC solution for Jenkins.

I know the team has thought through about that and @ndeloof even have some slides he used to discuss this matter with @ewelinawilkosz. I will work with them to surface that up in ways that Jenkins developers can access.

Also, I would note that since Daniel raised his original comment, more README was added to the repository that explains how to use this project in its current shape, which makes it a lot more accessible for others. Kudos for @ndeloof to get that work done!!

PR #43, opened 2017-12-28
[JEP-200] - Update the state of JEP-200 and accept it

oleg-nenashev:
This change reflects the current state:

- [x] Reflect the testing status. Lots of tests have been performed during the JEP review
- [x] Add notes from the discussion in the mailing list
- [x] Add https://wiki.jenkins.io/display/JENKINS/Plugins+affected+by+fix+for+JEP-200 to the requirements
- [x] Accep JEP

@jenkinsci/jep-editors @jglick @raul-arabaolaza @dwnusbaum

### DIFF
--- a/jep/200/README.adoc
+++ b/jep/200/README.adoc
@@ -414,3 +414,27 @@ According to the JEP sponsor, they were covered by their internal testing proced
 Although there will be extra testing performed before the release of the change in the Weekly,
 BDFL confirms that the current test coverage is good enough to accept this Jenkins Enhancement Proposal
 (Jan 08, 2017).
+
+=== Concern about the single whitelist approach
+
+https://github.com/oleg-nenashev[@oleg-nenashev] raised a concern about using the same whitelist for Remoting and XStream:
+
+* With the reference implementation in link:https://github.com/jenkinsci/jenkins/pull/3120[Jenkins PR #3120]
+there is no way to approve serialization only for a single serialization type
+(e.g. only XStream).
+* Possible attack vectors in Remoting and XStream differxw,
+especially when Remoting CLI is enabled due to the features missing in other CLI modes (multiple file parameters, etc.).
+In such case
+
+Feedback from the JEP Sponsor:
+
+* Remoting CLI is not a concern since we are going to consider it as insecure
+and deprecated option even after integrating
+* Current implementation can be extended in the future if needed.
+Jenkins core patch may be required to pass information about the serialization type to
+the `CustomClassFilter` implementations
+* `CustomClassFilter` extension point is restricted now,
+so any required adjustements can be made by API users when needed.
+
+The BDFL delegate agreed with the provided feedback (Jan 03, 2017).
+He would like to see better extensibility in the future, but it does not block delivery of JEP-200.
--- a/jep/200/README.adoc
+++ b/jep/200/README.adoc
@@ -257,6 +257,10 @@ in these cases plugin developers or administrators are likely to need instructio
 
 == Testing
 
+NOTE: This section is listed as described by the JEP Sponsor.
+Additional testing has been performed during the JEP-200 review in order to evaluate the proposal.
+Testing notes for the JEP review phase can be found in _Appendix A_ and the linked documents.
+
 The reference implementation includes test coverage for the essential aspects of the newly added filter:
 for example, that an example library class not currently included in the whitelist is rejected under the expected conditions.
 
--- a/jep/200/README.adoc
+++ b/jep/200/README.adoc
@@ -319,7 +319,7 @@ In such case new API will not be used, and hence there will be no behavior chang
 
 === Phase 2. Weekly Release
 
-Target date: Jan 06, 2017
+Target date: Jan 13, 2017
 
 0. Once this JEP is approved, the https://jenkins.io/redirect/class-filter/ will be created on the Jenkins website
 ** This document should provide a custom guide for creating JIRA issues with the `jep-200` label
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
I feel strongly that the proposed implementation should be tested against the standard Jenkins WAR. Acceptance Test Harness is mandatory IMHO, regarding PCT I am not 100% sure.

@daniel-beck @olivergondza what do you think as Security/LTS officers? What would be the preferable quality bar for that assuming we already have regressions in plugins confirmed by the JEP sponsor?

jep/200/README.adoc
```diff
@@ -319,7 +319,7 @@ In such case new API will not be used, and hence there will be no behavior chang
 
 === Phase 2. Weekly Release
 
-Target date: Jan 06, 2017
+Target date: Jan 13, 2017
```
jglick: Why?

jep/200/README.adoc
```diff
@@ -384,11 +384,18 @@ N/A
 
 Current state:
 
-* link:https://github.com/jenkinsci/acceptance-test-harness[Jenkins Acceptance Test Harness] has been executed with the patched components, several plugins were fixed (see the _Testing_ section)
-* link:https://github.com/jenkinsci/plugin-compat-tester[Plugin Compatibility Tester (PCT)] has been executed for a limited plugin set
+* link:https://github.com/jenkinsci/acceptance-test-harness[Jenkins Acceptance Test Harness (ATH)] has been executed with the patched components, several plugins were fixed (see the _Testing_ section)
+** After the discoveries in PCT (see below), the ATH results need to be reviewed carefully
+** BDFL Delegate is waiting for info about what has been tested
+(Jenkins WAR from link:https://github.com/jenkinsci/jenkins/pull/3120[Jenkins PR #3120] or a custom build with custom plugin versions)
```
jglick: Custom build using the source branches from that PR. Not sure what difference it makes.

jep/200/README.adoc
```diff
@@ -384,11 +384,18 @@ N/A
...
+* link:https://github.com/jenkinsci/plugin-compat-tester[Plugin Compatibility Tester (PCT)] has been executed for a limited plugin set for a sponsor's custom Jenkins WAR
 ** Not all plugins recommended in the Setup Wizard have been verified
+** The verified plugins have been checked with the sponsor's special set of plugins, not with the latest releases of plugins in Jenkins update centers.
+It cannot be considered as a reliable check since newer versions may contain other classes being serialized over Remoting/XStream.
```
jglick: Possible of course, but unlikely. The plugins are pretty close to current, and the problematic classes _tend_ to be ones which have been in use since “old times”.

jep/200/README.adoc
oleg-nenashev: Discussion summary:

* Some tested plugins were not at the latest version. In particular, Maven plugin is not 3.0
* Custom build alters Jenkins behavior a bit, but these changes should not impact XStream/Remoting serialization logic in tests

We will double-check which plugins need to be retested

oleg-nenashev:
@jglick PTAL

jep/200/README.adoc
```diff
@@ -24,9 +24,9 @@ endif::[]
 | Status
 // Uncomment the appropriate line.
 //| Not Submitted :information_source:
-| Draft :speech_balloon:
+// | Draft :speech_balloon:
 //| Deferred :hourglass:
-//| Accepted :ok_hand:
+| Accepted :ok_hand:
```
jglick: :tada:

jep/200/README.adoc
jglick: (resolved)

jep/200/README.adoc
```diff
@@ -378,29 +385,60 @@ so the maintainers will be the entry point.
...
+** It was decided that it is not enough (not recent plugin versions, potential impact on the plugin behavior by the custom logic),
+so the BDFL delegate and the JEP Sponsor re-run PCT with a standard Jenkins WAR
+** During testing the BDFL Delegate discovered issues which prevent him from running PCT in
+particular cases.
+These issues are listed in the link:https://issues.jenkins-ci.org/browse/JENKINS-48734[JENKINS-48734] EPIC.
+The blocker issues have been resolved.
+** All plugins recommended in the Jenkins Installation Wizard have been covered as well as many other plugins
+** BDFL delegate has not tested Pipeline and Blue Ocean plugins,
+because they are being maintained by employees of the JEP Sponsor's company.
+According to the JEP sponsor, they were covered by their internal testing procedure.
```
jglick: There is no “internal testing procedure” here, just

* regular ATH
* regular PCT (including all common Pipeline / Blue Ocean plugins)
* Blue Ocean’s regular acceptance tests

jep/200/README.adoc
```diff
@@ -378,29 +385,60 @@ so the maintainers will be the entry point.
...
+(Jan 08, 2017).
+
+=== Concern about the single whitelist approach
+
+https://github.com/oleg-nenashev[@oleg-nenashev] raised a concern about using the same whitelist for Remoting and XStream:
+
+* With the reference implementation in link:https://github.com/jenkinsci/jenkins/pull/3120[Jenkins PR #3120]
+there is no way to approve serialization only for a single serialization type
+(e.g. only XStream).
+* Possible attack vectors in Remoting and XStream differxw,
```
jglick: typo

jep/200/README.adoc
```diff
@@ -378,29 +385,60 @@ so the maintainers will be the entry point.
...
+especially when Remoting CLI is enabled due to the features missing in other CLI modes (multiple file parameters, etc.).
+In such case
```
jglick: …?

jep/200/README.adoc
jglick: FTR I do not agree that the possible attack vectors differ. Obviously non-`Serializable` classes cannot be exploited via Remoting, since they cannot be used at all.

jep/200/README.adoc
```diff
@@ -378,29 +385,60 @@ so the maintainers will be the entry point.
...
+
+Feedback from the JEP Sponsor:
+
+* Remoting CLI is not a concern since we are going to consider it as insecure
+and deprecated option even after integrating
+* Current implementation can be extended in the future if needed.
```
jglick: But I do not foresee any possible need for it in the future. This section misleadingly implies that _I_ think there is something here which could be improved but which we are deferring.

oleg-nenashev: Will make it explicit that there is no plan to work on it

jep/200/README.adoc
oleg-nenashev: OK, still sounds like an internal testing procedure since are results are not public. So I propose to consider it as a blackbox in this JEP

jep/200/README.adoc
oleg-nenashev: (resolved) ?

oleg-nenashev:
@jenkinsci/jep-editors I need an approval from somebody else to get it mergeable

jep/200/README.adoc
```diff
@@ -426,9 +426,11 @@ https://github.com/oleg-nenashev[@oleg-nenashev] raised a concern about using th
 * With the reference implementation in link:https://github.com/jenkinsci/jenkins/pull/3120[Jenkins PR #3120]
 there is no way to approve serialization only for a single serialization type
 (e.g. only XStream).
-* Possible attack vectors in Remoting and XStream differxw,
+* Possible attack vectors in Remoting and XStream differ,
 especially when Remoting CLI is enabled due to the features missing in other CLI modes (multiple file parameters, etc.).
-In such case
+** For attack via XML you usually need `Item.CONFIGURE` permissions
+** For attack over Remoting - `Computer.CONFIGURE` or write access to Remoting/Swarm Client JAR files on an agent.
+If Remoting CLI is enabled... then there is no special permissions required.
```
jglick: I see where you were going with this. Not attack vectors exactly, but preconditions for an attack.

oleg-nenashev:
I got the confirmation from the Security Team members that they are fine with the current state.
🚢 🇮🇹 

I am going to continue testing and reviews in the PRs, the specification in JEP is definitely fine.

PR #59, opened 2018-02-21
additional chapter on plugin management

ndeloof:


ndeloof:
ping @ewelinawilkosz  @MadsNielsen @buep @oleg-nenashev @jglick

oleg-nenashev:
I need more feedback from @jenkinsci/jep-editors about that. Maybe I misunderstand the idea for **informational** JEPs

bitwiseman:
@oleg-nenashev  
This is a Standards JEP, not informational.  Acceptance doesn't mean the document can't continue to change.  It means only that there is general agreement that this enhancement should and will be done and that the design as described is "acceptable".   If there are strong objections to notable design changes/additions (such as this one), we could return this JEP to Draft for reconsideration, but that would be an extreme case. 

At the same time...
@ndeloof and @ewelinawilkosz 
Oleg makes good point.  This is a rather sizable addition.  Do you foresee more additions on this scale happening?  It would be helpful to at least have a heading  for each feature area even if there is more work to be done to flesh out the design.  If there's enough items to be added, it might make sense to move them to a "CaC v2 JEP" (as Oleg suggests).

oleg-nenashev:
I would accept the approach for JEP extension if there is an explicit version history in the document directly (like other specifications do). I also see no problem with moving it back to Draft and editing if the JEP sponsor prefers it

bitwiseman:
@oleg-nenashev Version history makes sense.  I'll submit a PR to JEP 1 for version history.

bitwiseman:
@oleg-nenashev @ndeloof @ewelinawilkosz 
https://groups.google.com/d/topic/jenkinsci-dev/tw0ETwvboAM/discussion

bitwiseman:
@ewelinawilkosz @ndeloof 
Assuming #76 is accepted, it sounds like this JEP would need to be set to Draft status again.  I want to be clear that this would not be a step backwards for this JEP, but rather an accurate description of the current state of the JEP (design and scoping still in progress).  
Do you agree?

bitwiseman:
https://github.com/jenkinsci/jep/pull/79

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
 The latter would apply the configuration when updates are detected on file. It could benefit some way to lock down configuration for
 components configured by the configuration-as-code mechanism to be read-only on web UI.
 
+=== Plugin management
+As part of configuring Jenkins Master we support plugin installation and management. Configuration-as-Code plugin itself is out-of-scope, it has to be bundled with Jenkins so it can take control on master. But for all other plugins we want configuration-as-code to offer :
+
+- a high-level user friendly way to define "feature" plugins to be installed. Typically, end-user want to build github pull requests, he doesn't care this one will required `token-macro` for some low-level implementation reasons.
```
omehegan: "a high-level, user-friendly way..."
"Typically, end users want to..."
"GitHub"
"... they don't care that this will require..."

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+- a fully reproducible way to setup a master. Relying on `latest` for transitive dependencies is _not_ an acceptable solution. Upgrades should also be tracked as configuration changes.
+- ability to configure infrastructure : private update sites and http proxy to download plugins.
+
+==== Shrink wrap
+`npm` is a typical sample for a dependency manager which has to support comparable requirements : provide a reproducible dependency management in a moving javascript ecosystem. It supports a [`shrinkwrap` file](https://docs.npmjs.com/cli/shrinkwrap) to compute dependency list from a higher level definition, and be able to exactly reproduce this setup later.
```
omehegan: "a typical example of a..."
"provide reproducible dependency management..."
"to compute a dependency list from a higher-level definition..."

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+
+Configuration-as-Code plugin can offer plugin management _outside_ jenkins as a standalone java executable. It can then consume a `plugins.yaml` file, provided as part of the configuraiton, to resolve plugin dependencies and compute a full list of plugins to be installed, which would be stored as `plugins.shrinkwrap` file. This later file can be used to automate plugin installation.
```
omehegan: "The Configuration-as-Code plugin..."
"Jenkins"

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+
+==== Plugin installation
+Configuration-as-Code can consume this shrinkwrap file and compare the requested state _vs_ actual master installation. If some extra plugins have to be installed, it can then setup pluginManager and trigger plugins installation, then restart jenkins.
```
omehegan: "trigger plugin installation"
"Jenkins"

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+
+==== Shrink wrap
+`npm` is a typical sample for a dependency manager which has to support comparable requirements : provide a reproducible dependency management in a moving javascript ecosystem. It supports a [`shrinkwrap` file](https://docs.npmjs.com/cli/shrinkwrap) to compute dependency list from a higher level definition, and be able to exactly reproduce this setup later.
+
+Configuration-as-Code plugin can offer plugin management _outside_ jenkins as a standalone java executable. It can then consume a `plugins.yaml` file, provided as part of the configuraiton, to resolve plugin dependencies and compute a full list of plugins to be installed, which would be stored as `plugins.shrinkwrap` file. This later file can be used to automate plugin installation.
+
+==== Plugin installation
+Configuration-as-Code can consume this shrinkwrap file and compare the requested state _vs_ actual master installation. If some extra plugins have to be installed, it can then setup pluginManager and trigger plugins installation, then restart jenkins. 
+As the yaml configuration may have references to plugins components, this process has to be executed _before_ any other configuration.
```
omehegan: "plugin components"

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+
+For flexibility and convenience, `plugins.shrinkwrap` file is optional, and plugin installation would then just rely on PluginManager to manage transitive dependencies. Configuration-as-Code would then just log a warning about reproducibility.
```
omehegan: "the `plugins.shrinkwrap` file is optional"

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+
+==== Dependency management
+As `plugins.shrinkwrap` computation is executed outside Jenkins, we can offer an alternate implementation for transitive dependency management and dependency resolution. With extra metadata exposed in update center, we could offer some more advanced logic in dependency resolution, not just getting `latest` of all plugins. Such metadata would express known compatibility-breaks, synchronized releases for highly dependent plugins, recommended version, etc. This is typically what Linux package management system do.
```
omehegan: "in the Update Center..."
"package management systems"

ndeloof:
@oleg-nenashev JEP status has been clarified

jep/201/README.adoc
```diff
@@ -230,6 +230,28 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+The Configuration-as-Code can consume this shrinkwrap file and compare the requested state _vs_ actual master installation. If some extra plugins have to be installed, it can then setup pluginManager and trigger plugin installation, then restart Jenkins. 
+As the yaml configuration may have references to plugin components, this process has to be executed _before_ any other configuration.
+
+For flexibility and convenience, the `plugins.shrinkwrap` file is optional, and plugin installation would then just rely on PluginManager to manage transitive dependencies. Configuration-as-Code would then just log a warning about reproducibility.
+
+==== Dependency management
+As `plugins.shrinkwrap` computation is executed outside Jenkins, we can offer an alternate implementation for transitive dependency management and dependency resolution. With extra metadata exposed in the Update Center, we could offer some more advanced logic in dependency resolution, not just getting `latest` of all plugins. Such metadata would express known compatibility-breaks, synchronized releases for highly dependent plugins, recommended version, etc. This is typically what Linux package management systems do.
```
daniel-beck: Applying a fixed version configuration should check published security warnings from existing update-center.json and log warnings if the plugin configuration has warnings reported against it.

ndeloof: I agree it would be nice to include some security support here, anyway I'd prefer we don't extensively discuss what plugin management should look like (as noticed by @oleg-nenashev this also would apply to docker image). My major concern here is to offer plugin management with JCasC _and_ reproducibility.

I'd like we discuss plugin management v2.0 in a decidacted JEP. We need security updates, but also actual dependency management, compatibility breaks, recommended ... all sort of things system package manager have implemented but Jenkins doesn't offer. This should be packaged as a standalone tool we could use from JCasC but also docker image or bundled in jenkins.war.

jep/201/README.adoc
```diff
@@ -230,13 +234,40 @@ The former just uses the configuration-as-code mechanism for initial setup.
...
+
+
 
 == Backwards Compatibility
 
 Configuration-as-Code is intended to run as an additional Jenkins component (most probably: a plugin) and not require dedicated extension
 integrated in Jenkins-core nor specific API implemented by plugins. We only require them to follow some convention in the way they expose
 configuration attributes (i.e +DataBoundSetter|Constructor+)
 
+As Configuration-as-Code model is directly compute based on exposed API, this model might change with
```
bitwiseman: This is listed under backward compatibility, but it is as much about "Version Compatibility" (*Backward and Forward*).  This sounds like a potentially massive point point for users.  As such, this behavior - what doesCasC when the YAML doesn't match the version of Jenkins and/or it's plugins - needs to be more fully described as the customer is almost guaranteed to have to debug YAML config incompatibilities, possibly extensively and often.

ndeloof: Yes indeed, but this is the price to pay to avoid a new API to be implemented by 1500 plugins and many users to complain their favourite plugin isn't supported.

ndeloof:
closing this PR as we'd like plugin management and versioning being discussed at a higher level (this also impacts docker image, and probably plugin manager at some point)

PR #67, opened 2018-03-21
[JENKINS-49406] Evergreen snapshotting data safety system JEP

batmat:
[JENKINS-49406](https://issues.jenkins-ci.org/browse/JENKINS-49406)

### References
* [Draft PR](https://github.com/batmat/jep/pull/1) to process feedback from people on the dev ML 
* [Jenkins Essentials' JEP-300](https://github.com/jenkinsci/jep/tree/master/jep/300)
* [Jenkins Essentials' JEP-301](https://github.com/jenkinsci/jep/tree/master/jep/301)

@jenkinsci/jep-editors

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -67,47 +67,45 @@ endif::[]
 
 link:https://github.com/jenkinsci/jep/tree/master/jep/300:[Jenkins Essentials], and more specifically its link:https://github.com/jenkinsci/jep/tree/master/jep/301:[Evergreen] _component_, aims at providing an link:https://github.com/jenkinsci/jep/tree/master/jep/300#auto-update[automatically updating distribution] of Jenkins.
 
-Continuous Delivery is about making small incremental changes, making failures much more easily recoverable. In the context here, it means we need to be able to seamlessly upgrade Jenkins, _but also_ roll back to the previously running version if an upgrade goes wrong.
-As Jenkins does not support downgrading alone, we need to introduce a system that will enable that auto-downgrade capability.
-
-This document outlines the design of the snapshotting system we plan to use.
+Continuous Delivery is about making small incremental changes, making failures much more easily recoverable. In the context here, it means Jenkins Essentials must to be able to seamlessly upgrade Jenkins, _but also_ roll back to the previously running version if an upgrade goes wrong.
+As Jenkins does not support downgrading alone, this document introduces the snapshotting system that which enables that auto-downgrade capability.
 
 == Specification
 
-_Evergreen_ works with two main components: Jenkins itself, and what we call the evergreen-client.
+_Evergreen_ works with two main components: Jenkins itself, and the `evergreen-client`.
 
-=== Upgrading & downgrading
+=== Upgrading and downgrading
 
-Once an evergreen client has been instructed to perform an upgrade, it is responsible for the following operations:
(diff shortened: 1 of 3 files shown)


rtyler:
If no other JEP editor gets to this by later today PDT, i'll take a look once I have fresher eyes :smile:

jep/0000/README.adoc
```diff
@@ -0,0 +1,324 @@
...
+===== What to back up
+
+Thanks to the data segregation explained above, we will be snapshotting (almost) everything under `/var/jenkins/home`.
+
+We still need to have a `.gitignore` file for some things that either cannot be moved elsewhere, or that we do not want to store in the Git repository.
+As said above, this will likely be improved as we go.
+
+[source,gitignore,title=.gitignore]
+----
+plugins/
```
jglick: As I noted in the draft, this _may_ be overly aggressive. Probably you want to limit this to actual plugin binaries (not `*.disabled`) that are in the Essentials BOM:

```
plugins/structs.jpi
# etc.
```

batmat: For easier access, copying what I wrote there about this.
----
Jesse was saying:
> People may be expecting that a Git revision (plus master.key) corresponds to a fully reconstructible Jenkins instance.

Well, then they will be deceived... It's not a backup system as we already discussed.

To me that this part is going to be out of the scope of the snapshotting system, exactly for the reason you gave: I consider that the whole plugin list is defined explicitly by the BOM, for Essentials plugin.

I would say that for others, well it's also out of scope, and it will be the evergreen client responsibility to store the versions of **all** plugins present there, to be able to restore the previous version if we rollback.

samrocketman: In general, the expanded plugin data should go under `/var/cache/jenkins/plugins/` just like `/var/cache/jenkins/war/`.  This way one does not need to worry about catching the expanded plugin data.

samrocketman: You'll want to be explicit with you gitignore paths here.  For example, the following gitignore

```
plugins/
```
will ignore all subfolders whose name is plugins.  For example, lets say someone creates a job named `plugins` or a Jenkins folder named `plugins`.  The gitignore as-is will ignore those critical configurations.  gitignore applies to all subdirectories and not just the root.  Instead, use a full path relative to the root of the repository.  Example gitignore,

```
/plugins/
```

batmat: @samrocketman about the `/var/cache/...` comment can you please post back to the ML to check what/if people have to say about this?

I'm leaning towards changing it this way, but would like to follow the process first to avoid back and forth. Thanks

rtyler:
@jenkinsci/jep-editors With my JEP Editor hat on, I'm going to assign this a number, mark the status as `Draft` and merge while the discussion continues on the [mailing list](https://groups.google.com/d/msgid/jenkinsci-dev/CANWgJS7gojzJCpKZskJVz4e0toS_pAUvintduNRiPEog9532QA%40mail.gmail.com)

jep/0000/README.adoc
samrocketman: Sure, I don't mind.  It's how I configure my Jenkins installs.

samrocketman: Done https://groups.google.com/forum/m/#!topic/jenkinsci-dev/C8fBp3r98bY

PR #82, opened 2018-04-11
Require UTF-8 for simplicity

batmat:
JSON allows many possible encoding. We restrict it to UTF-8 only for simplicity.

>[JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
   encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
   interoperable in the sense that they will be read successfully by the
   maximum number of implementations; there are many implementations
   that cannot successfully read texts in other encodings (such as
   UTF-16 and UTF-32).](https://tools.ietf.org/html/rfc7159#section-8.1)


batmat:
@jenkinsci/jep-editors cc @rtyler

PR #83, opened 2018-04-13
External Artifact Storage

carlossg:


oleg-nenashev:
@jenkinsci/jep-editors

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+| Title
+| External Artifact Storage
+
+| Sponsor
+| https://github.com/carlossg[@carlossg]
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
```
jglick: should now be set to Draft IIUC (since that is what you are proposing)

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+The AWS S3 implementation tests exercise this flow plus add some extra S3 specific tests.
+They require an AWS account and S3 permissions and can be run from a EC2 instance or a local machine.
+Some mock testing can be added in the future.
+
+== Prototype Implementation
+
+https://github.com/jenkinsci/artifact-manager-s3-plugin[artifact-manager-s3-plugin]
+
+
+== References
```
oleg-nenashev: I also have a WAR packaging implementation for the reference implementation: https://github.com/oleg-nenashev/jenkins-custom-war-packager-ci-demo . Maybe makes sense to link it as well

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+Extra care needs to be taken so agents do not have any access to the blob store.
+
+In the common case where the vm instances are assigned roles (`IAM role` in AWS) the instance where the master runs should have access to the blob store but the agents should run in a different instance where its role does not allow it.
+
+In a Kubernetes environment this means either using different node pools for masters or agents or using something like https://github.com/jtblin/kube2iam[kube2iam] to have different roles per pod.
+
+
+== Infrastructure Requirements
+
+Ideally we could use Jenkins infrastructure to do live testing with S3, which is not currently possible due to lack of AWS account.
```
oleg-nenashev: Would be nice to briefly describe the fallback infrastructure you are going to use

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+== Motivation
+
+As mentioned in the abstract, running in a containerized environment is challenging due to the excessive use of the file system. Plus the lack of functionality provided by files vs dedicated artifact or blob store (versioning, permissions, cold store,...)
+
+From the usability point of view, currently a Jenkins installation sizing has to have into account the expected amount of disk needed or a restart, disk reprovisioning, disk swap,... instead of a infinite scale, pay per use that a blob store can offer. Cloud installations can automatically configure the blob store to use offering a much better first time user experience.
+
+Several alternatives exist today but they all require changes to all the pipelines and job definitions to explicitly choose the backend to send artifacts to.
+
+* https://plugins.jenkins.io/s3[S3 plugin]
```
oleg-nenashev: The plugins should be able to co-exist well if needed, right? Also CC @Jimilian

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+== Reasoning
+
+AWS is the focus as it is the most widely used cloud provider, S3 being the prevalent blob store.
+Equivalent features to S3 exist in other cloud providers and artifact repositories.
+
+The S3 implementation also uses http://jclouds.apache.org[Apache JClouds] that abstracts most of the implementation from the underlying blob store.
+
+== Backwards Compatibility
+
+Existing plugins using `ArtifactManager` API will continue to work using the new selected implementation.
```
oleg-nenashev: What about plugins NOT using this API? Will they continue working if this new Artifact Manager is disabled?

Would be also nice to provide a list of known plugins which work with artifacts directly without using this API (and hence will break in new impl). At the same point mention that the current Filesystem-based storage will continue working as before for that plugins

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+:warning-caption: :warning:
+endif::[]
+
+.**JEP Template**
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
```
oleg-nenashev: Number should be assigned

jep/0000/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Draft :speech_balloon:
+
+| Type
+| Standards
+
+| Created
+| 2018 Apr 11
```
oleg-nenashev: April 13

jep/0000/README.adoc
carlossg: yes, it wouldn't affect it

jep/0000/README.adoc
carlossg: done

jep/0000/README.adoc
carlossg: done

bitwiseman:
@oleg-nenashev The editor sets the number unless the sponsor requests a specific one.

jep/0000/README.adoc
bitwiseman: @oleg-nenashev The editor sets the number unless the sponsor requests a specific one.

jep/202/README.adoc
```diff
@@ -0,0 +1,230 @@
...
+This proposal provides the APIs needed to store all artifacts in an external location without going through the master.
+It includes an initial implementation using AWS S3.
+
+== Specification
+
+Jenkins agents upload and download artifacts directly to the external location using HTTP, no content traffic is sent over the remoting channel.
+
+All external store operations except upload/download are executed from the master, so the agent does not need any permissions other than HTTP upload/download. Pre-signed urls with expiration can be used so the agent does not have access to the full store.
+
+Stashes are stored as tarballs to ensure permissions and links are preserved.
```
batmat: is it implicit that we use `tar` only, or compressed tar (be it `.tar.gz`, `.tar.xz`...)?

If we do not want to compress it, because for instance we tested it and it's often unneeded, or even counter-productive, I would expect a part explaining this in the _Reasoning_ section below.

carlossg: tgz currently https://github.com/jenkinsci/artifact-manager-s3-plugin/blob/master/src/main/java/io/jenkins/plugins/artifact_manager_s3/JCloudsArtifactManager.java#L212

jglick: The default `stash` implementation using the master filesystem also uses `tgz`, but yes it is an implementation detail.

PR #84, opened 2018-04-13
Publishing incremental commits as Maven releases

jglick:
[JENKINS-50686](https://issues.jenkins-ci.org/browse/JENKINS-50686)

@jenkinsci/jep-editors

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,850 @@
+= JEP-0000: Publishing incremental commits as Maven releases
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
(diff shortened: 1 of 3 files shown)


jep/0000/README.adoc
```diff
@@ -0,0 +1,850 @@
...
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+
+== Abstract
+
+The Jenkins source base is spread across numerous GitHub repositories:
```
bitwiseman: Much of this abstract belongs in the motivation.  This section should be one paragraph that describe the proposal.   From what I can tell line 94 is about where the abstract could start, but it might be even line 99.  
Think of this as the elevator-pitch description of the proposed change.

PR #89, opened 2018-04-18
Sketching the Essentials update client/server lifecycle

rtyler:


jep/0000/README.adoc
```diff
@@ -0,0 +1,232 @@
...
+
+* `core` which indicates that a new jenkins.war is necessary.
+* `plugins` which will include a list of `updates` for plugins. This is an
+  object within the JSON structure rather than a flat array as it is expected
+  that at some point in the future we  may require a `removes` list to properly
+  unpublish legacy or out-dated plugins from instances.
+* `client` which incidates a new tarball for upgrading the `evergreen-client`
+  itself.
+
+Additional keys should be ignored by clients not supporting them to allow the
```
olblak: A warning in logs would be better than just ignoring unknowned  keys.
It helps to bring attention on something unusual

jep/0000/README.adoc
```diff
@@ -0,0 +1,232 @@
...
+----
+
+[[update-manifest]]
+=== Update Manifest
+
+The responses sent to the client must be well-formed JSON documents, referred
+to as "update manifests" which the client must understand. The Update Manifest
+should have a consistent structure but will be dynamically generated _per
+client_ in order to ensure that the client is only downloading what is
+necessary to update that specific client.
```
olblak: Do you consider all updates as 'safe'?
What happened if a client didn't connect to the update service for month?
Is it an information that would be useful in the update manifest?

jep/0000/README.adoc
rtyler: I'm not sure I understand your thoughts on a client-side log warning being useful here. I don't anticipate this being an unusual scenario but rather a scenario where the Update Service is providing more data than the current client knows how to handle, with the assumption that once the client is updated, it will know how to handle it.

jep/0000/README.adoc
rtyler: This is a really interesting question! Rephrased the problem would be "what if a client was offline for a longer period of time such as not to allow for a daily upgrade, what should our behavior be then."

I'll need to think about how we should handle that some more and consult with @batmat when he's returned tomorrow (Thursday).

My current guess is that we should try to update the client "one by one" through the update manifests over time, which has some data retention requirements I hadn't previously considered.


Thought provoking question for sure @olblak  :smile_cat:

jep/0000/README.adoc
```diff
@@ -0,0 +1,232 @@
...
+document specifies those API interactions and expectations to allow this update
+lifecycle to be performed correctly.
+
+== Specification
+
+The role of the versions service is simply to maintain an audit trail of sorts
+for the versions of sofware that each client has installed locally.
+
+
+The role of the update service is to use the `essentials.yaml` and retriev the
```
jglick: typo

jep/0000/README.adoc
```diff
@@ -0,0 +1,232 @@
...
+        "uri" : "https://updates.jenkins.io/some/path/to/a/jenkins.war",
+        "checksum" : {
+            "type" : "sha256",
+            "signature" : "somechecksumforthefile"
+        }
+    },
+    "plugins" : {
+        "updates" : [
+            {
+                "uri" : "https://updates.jenkins.io/some/path/to/a/plugin.hpi",
```
jglick: `updates.jenkins.io` as currently implemented cannot serve versioned content, much less commits from JEP-305. Can we serve from Artifactory (perhaps via repo.azure.jenkins.io proxy)?

rtyler: I'll replace with `example.com` domains for clarity. 

updates.jenkins.io is just an Apache server, so yes it can serve up whatever we put there :smiling_imp:

jep/0000/README.adoc
olblak: It's more a general concern about software that voluntary ignore information when it doesn't know how to handle it.

All information in a json should have a meaning especially when dynamically generated.
And if they don't, then they should be removed.

````
Example:
 "As a Jenkins Essential enduser, I expect all packages to be signed so that I am confident that my instance isn't corrupted or tampered"

 During an update, if this message pop up "WARNING: Unkowned key 'signature'"
 This brings valuable information.

 Package signature wasn't checked by the client because:
    1. There is a bug in my client version.
    2. My client version doesn't support this feature yet.
    3. My client version doesn't support this feature anymore.

This is totally different than hiding a key because the client doesn't support it.

So maybe 'warning' isn't an appropriate log level but this information should be accessible somewhere.

**Comment by rtyler** [comment]
@jenkinsci/jep-editors just an FYI, I consider this no longer work in progress and well-formed to be considered a Draft and numbered in the Jenkins Essentials series

File: jep/0000/README.adoc [hunk 5]
```diff
@@ -0,0 +1,469 @@
...
+== Specification
+
+There are two services involved behind the scenes which are responsible for
+facilitating the client's update lifecycle: `updates` and `versions`.
+
+The role of the versions service is simply to maintain an audit trail
+for the versions of software that each client has installed locally. As clients
+go through various upgrade processes, they will periodically `POST`
+<<version-manifest>> records which are stored in the Evergreen hosted service
+layer's storage.
```
batmat: Do we _tolerate_ to miss version updates? Or should the client always push *all* versions/update levels it went through? I guess the latter.

Is the following use case potentially to happen?

* Client Pushes to versions service that it is on UL 9 (Update Level)
* Client upgrades to UL 10
* Client upgrades to UL 11
* What does the client push to `versions` service ? That it _went_ through 10, and 11. Or only "hey, I am on UL11" ?

Sorry if this is dumb question. Maybe it will clarified below. Reading.

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+
+The role of the update service is to use the `essentials.yaml` and retrieve the
```
batmat: I guess we want a given `essentials.yaml` to be `==` to a given Update Level right? Also, I suppose we'll link to the https://github.com/jenkinsci/jep/pull/92 when ready to clarify what this is for newcomers.

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+must have the following data structure:
+
+* `id`: Numerical sequential integer to identify the update
+* `commit`: a Git commit corresponding to a change of `essentials.yaml` in the
+  link:https://github.com/jenkins-infra/evergreen[jenkins-infra/evergreen]
+  repository
+* `channel`: String denoting the <<channel>> for the update.
+* `manifest`: JSON formatted representation of `essentials.yaml` for ease of
+  storage and query in PostgreSQL
+* `tainted`: Boolean to indicate whether this update should be considered
```
batmat: I'm wondering if we should have a richer structure for _taintedness_. Is this really bad or good? I fear it might not be that simple. (I'm looking at instances that may hint a given UL is bad, when actually it's due to an _inessentials_ plugin that is installed specifically there)

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+=== Client Update Behavior
+
+The client must perform the necessary downloading of items referenced in the
+<<update-manifest>> and perform checksum validation before initiating a client
+update process. The exact sequence of events and what machinery must execute
+on the client is considered outside of the scope of this document.
+
+The client should also post a new <<version-manifest>> once an update lifecycle
+successfully completed to ensure that subsequent update check-ins result in
+accurate generated Update Manifest.
```
batmat: I think this answers my point above. 

But what happens if for some reason, backend is unreachable for some seconds or minutes after an upgrade? Do we lose this call, or should we store something locally to be able to send it fully when reconnection occurs?

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+[[version-manifest]]
+=== Version Manifest
+
+A version manifest is the symmetrically opposite of the <<update-manifest>> in
+that it should include the actual versions of software present on a Jenkins
+Essentials instance. This may include software which is outside of the update
+lifecycle.
+
+The purpose of the purpose of the version manifest is primarily for the client
+to report to the serve a fairly accurate state of the installed software in
```
batmat: `serve` => `server` typo

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+the instance.
+
+
+.Version Manifest
+[source,json]
+----
+{
+    "schema" : 1,
+    "container" : {
+        "commit" : "sha1 of the built container",
```
batmat: Image or container? Also, how do we store this inside the container, seems like a chicken-and-egg issue to me: adding a step in the Dockerfile to store a SHA would by definition add another layer?

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+            "node" : "output of node --version",
+            "npm" : "output of npm --version",
+            "java"  : "output of java -version"
+        }
+    },
+    "client" : {
+        "version" : "version of evergreen-client"
+    },
+    "jenkins" : {
+        "core" : "jenkins.awr embedded version",
```
batmat: `awr`  => `war`

jep/0000/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+which provides much more metadata to provide user-visible information.
+
+As Jenkins Essentials is intended to update automatically, the metadata
+(<<update-manifest>>), only needs to contain the URLs for packages and a
+checksum for validation. There are additional <<security>> concerns and
+reasoning discussed below.
+
+
+The Update Levels are a consideration to ensure that clients which have
+differeing levels of connectivity consistency can be safely updated.
```
batmat: `differeing` ? => differing ?

jep/0000/README.adoc
rtyler: Effectively yes

jep/0000/README.adoc
rtyler: Honestly, I think this is something we'll have to figure out once Jenkins Essentials is in the field. We can add additional columns in the future without difficulty, but I don't want to spend too much time hunting for ghosts right now.

PR #92, opened 2018-04-20
Bill of Materials

carlossg:
@jenkinsci/jep-editors 

https://groups.google.com/forum/#!topic/jenkinsci-dev/pR2ZQMj95Zc


jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+Essentials needs to enable selective downloads of core, bundled libraries and plugins for incremental upgrades.
+Pushing into maven repo is fine, although indexing time is big.
+Will need to mirror into Azure CDN at some point.
+
+The BoM needs to be translated to take advantage of existing infrastructure.
+
+* Essentials needs to generate a diff for upgrades and download the bits
+* Existing ATH uses a war binary that can be built from BoM
+
+Packages should be trackable back to a commit id, but during development the BoM should point to branches or tags, so we would have a BoM as input and a effective BoM as output with all the git refs resolved to specific commit ids.
```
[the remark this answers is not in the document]
jglick: > more useful is for it to refer to a `SNAPSHOT` version in that case

_Most_ useful would be for it to refer to a local file path, triggering a build from sources. I think the WAR packager has an option like this.

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+    # environments get added to the other section when enabled
+    - name: aws
+      plugins:
+        - groupId: org.acme
+          artifactId: acme2-plugin
+          ref: master
+          # version: 1.0
+      components: ...
+  # other sections can be added and ignored by default
+# the realized BoM after refs are resolved
```
rtyler: @carlossg I'm having trouble understanding this part of the format. Is this expected to all be within the same `yaml` file? Or would the "realized BoM" be a separate generated file output?

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+    artifactId: ...
+    version: ...
+  annotations:
+    io.jenkins.x: y
+spec:
+  core:
+    # version OR version + ref (version just to keep Maven happy about version order)
+    ref: master
+    version: 1.0
+  components:
```
rtyler: While `core` and `plugins` are fairly easily understood, I'm not sure what `components` conceptually maps too. Perhaps it would be helpful to explain these root keys and their purpose below the sample file.

samrocketman: Is a component like a community pipeline shared library?  Basically, any code which doesn't fit into the norms of typical Jenkins contributions.

oleg-nenashev: In my cases "components" could be JARs and System Groovy Scripts to be injected. But I agree, it needs to be documented

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+This proposal aims to add new functionality and reuse existing tooling by generating Maven poms and other formats in use today.
+
+== Security
+
+There are no security risks related to this proposal.
+
+== Infrastructure Requirements
+
+There are no new infrastructure requirements related to this proposal.
```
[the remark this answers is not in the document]
oleg-nenashev: IMHO we want to create a step in pipeline-library, e.g. via integration with Custom WAR Packager. After that BOM will be accessible to all users within Jenkins infra if needed

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+Core, libraries and plugins could have a BoM in the master branch.
+This BoM should point to master of core, master of libraries.
+There is an implicit dependency to the current branch of the current git repository.
+
+The BoM should be processed to update the Maven pom on change events.
+
+==== Changes to Core
+
+Developer needs some changes to core or Jenkins internal library or component for downstream use.
```
oleg-nenashev: What exactly do you want to support in BOM? Some components like Extras Executable WAR have non-trivial packaging logic, which cannot be easily integrated into Core build. Without https://issues.jenkins-ci.org/browse/JENKINS-48578 it is not possible to patch such component easily, and likely the rebuild is required anyway

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+== Specification
+
+=== What exists today
+
+* https://github.com/jenkinsci/acceptance-test-harness[Jenkins acceptance test harness]
+* https://github.com/jenkinsci/blueocean-acceptance-test[Blue Ocean acceptance test]
+* https://github.com/jenkinsci/plugin-compat-tester[Jenkins plugin compat tester]
+* https://github.com/jenkins-infra/evergreen[Essentials evergreen draft]
+* https://github.com/jenkinsci/jenkins-war-packager[Jenkins WAR packager] recently started
```
oleg-nenashev: Wrong link. https://github.com/jenkinsci/custom-war-packager

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+* Reduce code review overheads so that maintainers can process more.
+AKA “I am a plugin maintainer, not a FindBugs robot. Automate checks for me”
+
+
+== Specification
+
+=== What exists today
+
+* https://github.com/jenkinsci/acceptance-test-harness[Jenkins acceptance test harness]
+* https://github.com/jenkinsci/blueocean-acceptance-test[Blue Ocean acceptance test]
```
oleg-nenashev: Not related to packaging. 
For packaging I would rather reference https://github.com/jenkinsci/docker

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+There are no testing issues related to this proposal.
+
+== References
+
+* https://github.com/jenkinsci/acceptance-test-harness[Jenkins acceptance test harness]
+* https://github.com/jenkinsci/blueocean-acceptance-test[Blue Ocean acceptance test]
+* https://github.com/jenkinsci/plugin-compat-tester[Jenkins plugin compat tester]
+* https://github.com/jenkins-infra/evergreen[Essentials evergreen draft]
+* https://github.com/jenkinsci/jenkins-war-packager[Jenkins WAR packager]
```
oleg-nenashev: Wrong link again

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+//
+//
+// Uncomment if there will be a BDFL delegate for this JEP.
+//| BDFL-Delegate
+//| :bulb: Link to github user page :bulb:
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
```
oleg-nenashev: Needs to be updated. https://groups.google.com/forum/#!topic/jenkinsci-dev/pR2ZQMj95Zc

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+== Infrastructure Requirements
+
+There are no new infrastructure requirements related to this proposal.
+
+== Testing
+
+There are no testing issues related to this proposal.
+
+== References
```
oleg-nenashev: https://github.com/jenkinsci/jep/tree/master/jep/1#required-sections . "Prototype Implementation" section is missing. Without that JEP cannot be accepted as draft. My proposal would be to create a prototype implementation in https://github.com/jenkinsci/custom-war-packager.

jglick:
I think we have a fundamental clash of vision for the development workflow. I will try to better explain my perspective.

> the goal is to free developers from tedious tasks

That is a reasonable constraint, but it does not have to be at the expense of deterministic, reproducible builds.

My acid test would be this: if `/job/my-plugin/job/master/lastBuild/` is blue, and I

```bash
echo >> README.md
```

and file a PR from that, I expect `/job/my-plugin/job/PR-123/1/` to also be blue, since I could not possibly have broken anything by doing this. (Unless my build process somehow process data out of the README or something like that, of course.)

If the trivial PR is filed by the plugin maintainer, _maybe_ we could tolerate a failure unrelated to the actual contents of the PR, if you can give up the benefits of `git bisect` and so on. But if it was filed by an outside contributor, this is really bad. They are being given negative feedback about something which was not their fault, and which they probably do not even comprehend.

> In my plugin I want to point to core master and automatically keep retesting my plugin when core is updated, I want to know if something in core breaks it, without having to update versions continuously

Pointing to a nondeterministic branch of core in a BOM file in a plugin is neither necessary for this goal, nor sufficient.

First, it is not sufficient since you would only notice a problem when a new build happens to run on one of the branch projects, either due to a change in `master` or some new or updated PR. But if the plugin is mostly dormant, months could elapse between such changes, during which breaking core changes could have been “promoted” to weeklies or even LTS releases. It is also not sufficient because the specific tests requested in this plugin’s BOM are only those which you _expect_ are most likely to be broken, based on intuition and historical experience—and would be _downstream_ tests, generally unaffected by changes upstream to _this_ repository.

Second, it is not necessary. We must anyway have a CI job running on some schedule (say, nightly) that runs _all_ tests against `master` of _all_ components, and some team/process for triaging the failures and assigning them to responsible developers for detailed investigation and correction. In the context of Essentials, we must also be running all stable tests of “essential” components when a new deployment is proposed, which would also reveal any integration problems that had not been caught by an earlier test phase.

In either case, after the immediate problem is corrected (fixing a regression, adapting to it, or correcting a faulty test), we should take two follow-up steps:

1. Bump the upstream version number in the downstream BOM, so that the downstream repository is now testing against a newer baseline. This should of course be done in a PR to verify that everything passes before merging.
   (Cf. [JENKINS-47498](https://issues.jenkins-ci.org/browse/JENKINS-47498): ideally, this could be as simple as incrementing some integer which selects an entire Essentials BOM for import of versions. In other words, a downstream BOM should not need to exhaustively enumerate dozens of individual component versions—just an overall baseline version, plus some overrides corresponding to upstream PRs being consumed.)
1. Encourage developers to improve upon the late discovery of the problem by reproducing the bug in a more unit-like test in a more upstream repository, to pay down technical debt.

> If my plugin PR depends on a PR in core I want to point to that PR and automatically keep retesting my plugin PR when the core PR is updated

This is better done with a script which allows you to both consume the updated upstream PR and make any matching downstream adaptations in a single _atomic_ Git commit: [JENKINS-50699](https://issues.jenkins-ci.org/browse/JENKINS-50699).

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+* https://github.com/jenkinsci/acceptance-test-harness[Jenkins acceptance test harness]
+* https://github.com/jenkinsci/blueocean-acceptance-test[Blue Ocean acceptance test]
+* https://github.com/jenkinsci/plugin-compat-tester[Jenkins plugin compat tester]
+* https://github.com/jenkins-infra/evergreen[Essentials evergreen draft]
+* https://github.com/jenkinsci/jenkins-war-packager[Jenkins WAR packager] recently started
+
+=== What are we missing
+
+* Builds/Tests from master, needed for Essentials
```
batmat: Shouldn't this be moved below somehow under the _Testing_ section, where it currently says that there's nothing in here about Testing?

jep/0000/README.adoc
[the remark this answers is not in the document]
batmat: I do like the idea of freeing developers from tedious tasks, but as much as I like that idea, I am not sure we here have a clear audit trail/traceability of what _versions_ exactly go to production. And I do deeply dislike the idea of "version range" or something similar. And saying one uses a given branch is similar to this: you don't really know what you're using, and it can change somewhere.

I prefer the _GitOps_ approach described in JEP-305 where things are explicitly updated using a PR for instance. I think we must design things for end-user consumption, _then_ we can think about how to create tooling to make things easier to bump/update locally for quick development turnaround.

_Maybe_ here what you have in mind is not contradictory to this, but I'm just voicing my concern in case.

carlossg:
Curruntly travelling. Will do in the following week or so

jep/0000/README.adoc
rtyler: Related to my comment in [JENKINS-50953](https://issues.jenkins-ci.org/browse/JENKINS-50953), I believe the format for "realized" should contain referenceable URLs, This will make this file digestable and immediately usable by the Jenkins Essentials backend tooling.

oleg-nenashev:
I have created [JENKINS-51077](https://issues.jenkins-ci.org/browse/JENKINS-51077) for BOM support in Custom WAR packager and started working on the reference implementation

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+    version: 1.0
+  components:
+    - groupId: org.acme
+      artifactId: acme-component
+      ref: master
+      version: 1.0
+  plugins:
+    - groupId: org.acme
+      artifactId: acme-plugin
+      ref: master
```
oleg-nenashev: Git's remote is not defined here, we cannot easily determine it for plugins NOT hosted in update center

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+    - name: aws
+      plugins:
+        - groupId: org.acme
+          artifactId: acme2-plugin
+          ref: master
+          # version: 1.0
+      components: ...
+  # other sections can be added and ignored by default
+# the realized BoM after refs are resolved
+status:
```
jglick: What is `status`?

jep/0000/README.adoc
```diff
@@ -0,0 +1,264 @@
...
+  components:
+    - groupId: org.acme
+      artifactId: acme-component
+      ref: master
+      version: 1.0
+  plugins:
+    - groupId: org.acme
+      artifactId: acme-plugin
+      ref: master
+      # version: 1.0
```
jglick: After some work on [JENKINS-50953](https://issues.jenkins-ci.org/browse/JENKINS-50953), I have a suggestion here.

First, ditch the concept of “realized BOM”. This is an aspect of my broader comment made earlier about deterministic builds. So for sake of argument, let us say there is just one BOM format, and it pins particular versions of every component.

Let us also say that the version is simply a Maven version string, as opposed to the URLs suggested by @rtyler which seems overdetermined:
* makes it harder to apply caches and proxies
* for any given product build we already have a well-known list of repositories to look in
* the update tool needs to know that same list of repositories anyway
* more verbose and error-prone

Then you could have several variants. For a plain old trunk release from `maven-release-plugin` (“MRP”):

```yaml
- groupId: io.jenkins.plugins
  artifactId: stuff
  version: 1.23
```

There is an implied `branch: master` attribute here which is elided for brevity. For an incremental release from trunk:

```yaml
- groupId: io.jenkins.plugins
  artifactId: stuff
  version: 1.24-rc9999.abc123def456
```

For an incremental release from an _unmerged_ PR:

```yaml
- groupId: io.jenkins.plugins
  artifactId: stuff
  version: 1.24-rc9999.abc123def456
  branch: jglick:refactoring-JENKINS-56789
```

(Note the syntax matching [GitHub conventions](https://developer.github.com/v3/repos/commits/#compare-two-commits).) Even other options are possible, like MRP from a backport branch (in the origin repo):

```yaml
- groupId: io.jenkins.plugins
  artifactId: stuff
  version: 1.23.1
  branch: stable
```

Wait, you exclaim! If we are pinning a `version`, why do we need a `branch` too? Two reasons.

First, the version update tool needs to know which branch to inspect for candidate versions. Remember that while incremental version numbers are meaningfully comparable within a branch, that is not true between branches, since commit history can be a DAG, so it cannot simply pick the deployed release with the highest `$(git rev-list --first-parent --count HEAD)`; it has to filter them by branch. If we want to be able to run the tool in a mode where it offers updates to every component in the file without further ado, it needs to know if there are some which are currently being taken from something other than `master`.

Related to that, we ought to be performing automated validation if this BOM is being used for production, such as the Essentials manifest: a CI build of the BOM repo could use a subset of the code in JENKINS-50953 to check that a requested incremental `version` indeed maps to something already pushed to the designated branch. As alluded to in JEP-305, this is important for security. Suppose you receive a pull request that says

```diff
-  version: 2.13-rc735.f7ae668c009b
+  version: 2.13-rc737.789e66b233dd
```

Sure, that looks OK, right? A plugin update, go for it. But how do you know what this `789e66b233dd` commit actually was? Something you trust? Are you going to construct the GitHub `commit` URL for it and take a look? Probably not, you will just merge and forget. But that might have been malware built from some fly-by-night PR which was closed the moment its artifacts got deployed to Incrementals! By contrast, with the `branch` system that PR would be marked as a build failure, since the commit is not an ancestor of the current `master`, and if the pull request looks like

```diff
-  version: 2.13-rc735.f7ae668c009b
+  version: 2.13-rc737.789e66b233dd
+  branch: l33thaxor:muahaha
```

you are going to think twice before merging it.

(A particular BOM like the Essentials manifest may of course wish to use only `master` commits as a matter of policy, and it might also be receiving routine updates from a bot rather than PRs filed by actual people, but it would be best if there is a single format that can be used safely in different contexts.)

jglick:
Pulling a conversation out of a line item since I think it is a pretty basic question.

> The BoM is the source of truth and the tooling would need to update the pom to use the right artifact version (the commit id)

There is no concrete proposal here for how that would happen, or discussion of how that would affect existing developer workflows.

My tentative conclusion from what seems to be the first aspects of a reference implementation in https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/20#discussion_r186496426 is that we do not need a general BOM format at all. The custom WAR packager needs a small amount of configuration, and Evergreen’s config file is needed to define what goes into production, but I do not see any need for other tools to interact with either of these. It _would_ be useful for the `evergreen` repo to generate and deploy a Maven BOM (a `pom.xml` with only `<dependencyManagement>`) that we could import from other POMs to set defaults for artifact versions: [JENKINS-47498](https://issues.jenkins-ci.org/browse/JENKINS-47498).

oleg-nenashev:
In https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/20 we would likely need to generate BOM from a BOM Template and `pom.xml` to make it fancy from user PoV. Integration tests for Essentials components unlikely need BOM, because Custom WAR Packager and Pipeline methods do the job (and it's generally more flexible). But BOM may be still useful as unified format.

From Testing perspective, I am using BOM reference implementation in some internal flow зкщещензуы (actually, only the `status` section), but I could replace it if needed. CWP lacks support of `environment`s, but it can be added if needed.

jglick:
One thing I can see a BOM adding is a source branch override (replacing `master`, as in https://github.com/jenkinsci/jep/pull/92#discussion_r185943025) for those dependencies taken from unmerged PRs: the 4.0.0 POM schema does not offer any good place to add custom metadata. On the other hand you could just override [this mojo configuration](https://github.com/jglick/incrementals-tools/blob/a55c668d197d863b8605007b749991b992369d7a/maven-plugin/src/main/java/io/jenkins/tools/incrementals/maven/UpdateMojo.java#L67-L72) in your PR’s patch to `pom.xml`; for Evergreen security validation we can just assume `master`.

bitwiseman:
This JEP is **still** in a pre-draft state?

@oleg-nenashev said this couldn't be approved as draft until he implemented a prototype. That has been done.  I've cleared his request for changes. 

@carlossg @oleg-nenashev @jglick @jenkinsci/jep-editors 
Is there any reason why this shouldn't be set to draft?

jep/0000/README.adoc
bitwiseman: > Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.

oleg-nenashev:
After the prototype we started the discussion whether it is needed, and so far there is no response from the JEP Sponsor. IMHO JEP approval as draft was premature

oleg-nenashev:
It's not a big deal, it's always possible to return it to draft or reject later

bitwiseman:
@oleg-nenashev 
While it is fine for JEPs to remain in draft state for an extended period, it is not great for them to remain in "submitted" state for very long.  As you say, the JEP can be reject later if needed.

PR #96, opened 2018-04-20
[on-hold] Accepting JEP-202

jglick:
@kohsuke says
> I've assigned Jesse as BDFL Delegate to this JEP.

and

> I also think the amount of coding effort that went into this is well past the criteria that I expect for "Accepted"

BTW the `set-jep-status` script does not work for me:

```
jep$ (cd jep; ./set-jep-status 202 Accepted)
sed: can't read s/^| Draft :speech_balloon:/| Accepted :ok_hand:/: No such file or directory
```

CC @carlossg @rtyler


bitwiseman:
@jglick 

Thanks for the full and reasoned response.  Thanks for taking the time to document your reasoning fully.  It addresses my concerns about giving proper consideration to moving to Accepted. 

Let me make sure the context of my comments are clear.  We just had JEP have to go from Accepted to Draft due to it being moved to Accepted before it was ready.  I'm also confused as to why you as Reviewer are pushing to move this JEP to Accepted status. 

As Reviewer, it's not your job to make the case for this JEP to be Accepted.  Your job is to review the JEP from a critical stand point and consider ways in might _not_ be ready for "Accepted" status. I understand that you are responding to me because this is your PR, but it should really be Carlos making the request, making the case, and responding to my concerns.  

Further, since there are concerns voiced by a member of the community (me), that means that this JEP has not reached consensus. One of the concern I had that you did not really address was that "Specification section of the JEP seems extremely light and vague". As Reviewer in the JEP process, your response this kind of concern needs to be "I see you have concerns. That is enough reason for us to wait on Acceptance until the Sponsor, @carlossg, can address them." Then @carlossg and I would figure out what changes if any need to be made.

oleg-nenashev:
> Not sure exactly what that means but I suppose it is “stable”. Nothing that I know of has been proposed which would materially change the current JEP. @oleg-nenashev suggested that an appendix be added with a list of plugins known to not honor the ArtifactManager / VirtualFile API, which I plan to write soon.

I would vote for doing it before accepting JEP

Since the time-critical part for the core is integrated with `@Beta` status, IIUC we have some time to process this JEP. I would be really interested to have this API in final state by the next LTS baseline, but we have few months for that.

> @oleg-nenashev sounded like he might have more to say about the design after the JEP was approved as Draft. Has the sponsor had time to check back with him to make sure his feedback is

* The JEP does not specify Core API changes, and this is the most important part. "This change adds more methods to the VirtualFile API to support this flow" is not a specification at all
* JEP text still does not explicitly say that there will be no changes for instances which use old filesystem Artifact Manager implementation. From the JEP text it's not clear whether it is in the testing scope
* I am still concerned that API does not support `InterruptedException` in new and old methods (https://github.com/jenkinsci/jenkins/pull/3302). IMHO it may cause integration issues with plugins like Build Timeout plugins which may want to interrupt the archiving operations. Not a blocker since old APIs didn't do that, but it worths discussion. 
* As a potential plugin user, I am concerned that the reference implementation in S3 Artifact Manager does not persist references to buckets/folders within `build.xml`. Reconfiguration of artifact manager plugin may lead to the data unavailability and to malfunction of `BuildDiscarder` logic. Since it is a reference implementation, I do not see it as a blocker for accepting JEP. The Factory-based architecture itself supports injecting metadata into runs.

bitwiseman:
@carlossg 

We're all new to this process.  I'm not trying to criticize you.   If you'd like, I'd be happy to schedule a some time to chat about it and make sure you have all the information you need. 

> These conversations are not public because the origin of the plugin is not OSS but that explains why he has (almost) no concerns about what's stated in the JEP.

Some form of those conversations and/or the resulting decisions needs be covered by the JEP. 

> I have separately contacted a couple of possibly interested downstream consumers (cloud providers/repository providers) and will reach to more. IMHO that's when the useful actionable feedback is going to be provided and we now have the @Beta mechanism to allow exactly that.

This is a perfect summation of why I was suggesting to @jglick that this JEP is not ready for acceptance. Both of you are saying "We're still gathering feedback that has the potential to change the JEP and the design."   Between this and the concerns @oleg-nenashev voiced, I don't see how this JEP could be considered ready for "Accepted" status.  

And let me say again, there is no reason to rush a JEP to "Accepted" status.  There is absolutely no advantage.  (The only case where it could matter is if there were another JEP in competition with this one, which doesn't apply here at all.)

bitwiseman:
@carlossg 
Do you remember who indicated that this JEP needed to be accepted in order for APIs to be published?  That is deep misunderstanding that I want to make sure is cleared up.

jglick:
> I would vote for \[adding a list of affected plugins\] before accepting JEP

OK, can wait for that if you prefer.

> The JEP does not specify Core API changes

Probably it should link to [`VirtualFile` Javadoc](http://javadoc.jenkins.io/jenkins/util/VirtualFile.html) (unfortunately not yet rebuilt??) with a list of new methods; as well as `StashAwareArtifactManager` (not yet merged and so not yet available as online Javadoc, but could perhaps link to [sources](https://github.com/jglick/workflow-api-plugin/blob/44bc9985b0aa1fb9407b5c78be6b429ad29af2d8/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java#L292-L315) in the meantime).

> I am still concerned that API does not support `InterruptedException` \[…which\] may cause integration issues with plugins like Build Timeout plugins

As discussed in the core PR, I do not see any evidence that it would ([JENKINS-50597](https://issues.jenkins-ci.org/browse/JENKINS-50597), planned for work soon, would verify this), but at any rate the explanation does deserve to be in the **Reasoning** section.

Also @bitwiseman alerted me to the fact that the discussion of why `asRemotable` was ultimately removed from the API in favor of solely using `toExternalURL` should be moved to the **Reasoning** section, with the **Security** section only documenting the final state, to follow the standard JEP structure.

> I am concerned that the reference implementation in S3 Artifact Manager does not persist references to buckets/folders

It could; there are reasons we declined to do so. Could be discussed under **Reasoning**, though it is really a policy decision for this particular implementation—I do not think it is essential for readers of the JEP.

carlossg:
Added more details in #97

bitwiseman:
https://github.com/jenkinsci/jep/pull/117

bitwiseman:
Closing this for now.  Please reopen when this is ready for review.

jglick:
This should be mergeable once #117 is. I seem unable to reopen it at the moment.

bitwiseman:
Per reviewer @jglick, I'm setting this to accepted.

PR #98, opened 2018-04-26
Fleshing out backwards compatibility section of JEP-202

jglick:
As suggested by @oleg-nenashev in #96, enumerating plugins known to be using APIs which would clash with remote artifact storage. CC @carlossg

### DIFF
--- a/jep/202/README.adoc
+++ b/jep/202/README.adoc
@@ -154,16 +154,76 @@ The S3 implementation also uses http://jclouds.apache.org[Apache JClouds] that a
 == Backwards Compatibility
 
 Existing plugins using `ArtifactManager` API will continue to work using the new selected implementation.
+However, there are two classes of potential incompatibility.
 
-However:
+=== File-oriented artifact reference
 
-* Various plugins call `Run.getArtifactsDir` and similar deprecated APIs.
-These would already have been broken for users of the Compress Artifacts plugin, but that is rarely used, whereas we are proposing lots of people run with the S3 artifact manager.
-Calls to the deprecated APIs will behave as if there were no artifacts in the build.
+Various plugins call deprecated APIs which assume that build artifacts are stored as files inside the master’s build directory.
+These would already have been broken for users of the _Compress Artifacts_ plugin, but that is rarely used, whereas we are proposing lots of people run with the S3 artifact manager.
 We could add telemetry so that such calls produce a warning in the system log, at least when the build actually does have a custom artifact manager selected.
 
-* Some plugins using `VirtualFile` may still be calling open and then passing the stream to an agent.
+As seen in
+link:https://ci.jenkins.io/job/Infra/job/deprecated-usage-in-plugins/job/master/lastSuccessfulBuild/artifact/output/usage-by-api.html#hudson_model_Run_getArtifactsDir__Ljava_io_File_[this report],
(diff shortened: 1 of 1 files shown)


jep/202/README.adoc
```diff
@@ -154,16 +154,76 @@ The S3 implementation also uses http://jclouds.apache.org[Apache JClouds] that a
...
-* Some plugins using `VirtualFile` may still be calling open and then passing the stream to an agent.
+As seen in
+link:https://ci.jenkins.io/job/Infra/job/deprecated-usage-in-plugins/job/master/lastSuccessfulBuild/artifact/output/usage-by-api.html#hudson_model_Run_getArtifactsDir__Ljava_io_File_[this report],
+there are a number of plugins on the usual update center still calling `Run.getArtifactsDir()` and/or `Run.Artifact.getFile()`,
+despite the fact that these methods were deprecated in Jenkins 1.531 in 2013 as part of
+link:https://issues.jenkins-ci.org/browse/JENKINS-17236[JENKINS-17236].
+These include:
+
+[cols="5,>",options="header",width="50%"]
+|============================
```
oleg-nenashev: Would it make sense to create RFE tickets and reference them from JEP? maybe it's rather preferable to it in a Wiki page so that we could do it dynamically (like  https://wiki.jenkins.io/display/JENKINS/Plugins+affected+by+fix+for+JEP-200 or https://wiki.jenkins.io/display/JENKINS/Plugins+affected+by+fix+for+SECURITY-170)

Then the warning message for old API could redirect people to that page

jep/202/README.adoc
```diff
@@ -154,16 +154,76 @@ The S3 implementation also uses http://jclouds.apache.org[Apache JClouds] that a
...
-Calls to the deprecated APIs will behave as if there were no artifacts in the build.
+Various plugins call deprecated APIs which assume that build artifacts are stored as files inside the master’s build directory.
+These would already have been broken for users of the _Compress Artifacts_ plugin, but that is rarely used, whereas we are proposing lots of people run with the S3 artifact manager.
 We could add telemetry so that such calls produce a warning in the system log, at least when the build actually does have a custom artifact manager selected.
 
-* Some plugins using `VirtualFile` may still be calling open and then passing the stream to an agent.
+As seen in
+link:https://ci.jenkins.io/job/Infra/job/deprecated-usage-in-plugins/job/master/lastSuccessfulBuild/artifact/output/usage-by-api.html#hudson_model_Run_getArtifactsDir__Ljava_io_File_[this report],
+there are a number of plugins on the usual update center still calling `Run.getArtifactsDir()` and/or `Run.Artifact.getFile()`,
+despite the fact that these methods were deprecated in Jenkins 1.531 in 2013 as part of
```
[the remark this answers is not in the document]
jglick: Yes, was already suggested in this JEP, but does not do much good until JEP-304 is live I suppose.

jep/202/README.adoc
jglick: I did not consider it worth bothering given the limited impact and the low installation counts. Would be possible if adding core logging, of course.

At any rate, that does not seem to have much bearing on the text of the JEP.

oleg-nenashev:
I didn't merge it because there was no JEP Sponsor's approval at that point.
Merging it now

PR #101, opened 2018-05-07
YAML syntax fix

jglick:


oleg-nenashev:
Needs confirmation from the JEP sponsor. Likely it's an overkill in this case tho

PR #102, opened 2018-05-07
Moved user guide

jglick:
After [HOSTING-556](https://issues.jenkins-ci.org/browse/HOSTING-556) this is the more natural home.


oleg-nenashev:
It's definitely reasonable since there is similar tooling for https://github.com/jenkinsci/pom

PR #100, opened 2018-05-03
[JEP-5] - Jenkins Ambassador Program

bitwiseman:
@alyssat and I are working on getting the Jenkins Ambassador program going.  Starting this PR to gather early feedback.

PR #103, opened 2018-05-08
[JEP-1] - Add status update to the "Approve as draft" process

oleg-nenashev:
@jenkinsci/jep-editors


oleg-nenashev:
Actually in #100 I firstly updated JEP in @bitwiseman's, then pushed the changes to the PR and then squashed everything in order to avoid changes outside pull requests. Not sure what would be preferable.

bitwiseman:
@oleg-nenashev Thanks!

PR #106, opened 2018-05-15
clarification

ndeloof:


rtyler:
@ewelinawilkosz I perhaps merged this too soon without you having an opportunity to review it. My understanding is that @ndeloof's updates are based on the discussion in [this thread](https://groups.google.com/d/msgid/jenkinsci-dev/20180510201923.GF2935%40grape.lasagna.io).

If you wish I can revert.

ewelinawilkosz:
no need, I followed the discussion and I agree with Nicolas' updates
thanks anyway :)

bitwiseman:
@ndeloof @ewelinawilkosz 
Speaking as contributor (not a JEP Editor),  I do feel this was merged too quickly as it does not accurately represent the feedback in https://groups.google.com/d/topic/jenkinsci-dev/nnfreZqwIJ4/discussion  (it also doesn't reference that thread). 

Re: [line 260](https://github.com/jenkinsci/jep/commit/adfe51ad87ac1870affbaf0d0ecbcc9f8418c41a#diff-d37718df6e1dbf6f51cbde624d32e737R260):

> Configuration-as-Code doesn't define a model for the configuration expressed in Yaml file. The yaml schema will depend on a specific versions of jenkins-core and plugins being used at runtime. Any change to a plugin will change this model, this will happen any time a `DataBoundConstructor` is modified, setter added or renamed, etc. This is the price to pay for alignment with UI databinding and support for any plugin out of the box ("no glue code").

This is not the only possible way to address the complexity of myriad plugins. Saying "This is the price to pay" is not sufficient. For example, the CasC plugin could generate a schema file for a particular configuration and inform the user when that schema changes in a breaking way. I'm sure there are reasons you do not want to do this, but you cannot simply assert that there is no other way here - there needs to be more discussion of the choices.

Re: [line 266](https://github.com/jenkinsci/jep/commit/adfe51ad87ac1870affbaf0d0ecbcc9f8418c41a#diff-d37718df6e1dbf6f51cbde624d32e737R266):

> For the same reason, Configuration-as-Code way to infer the configuration model from a live instance will depend on Configuration-as-Code plugin version being used, and the resulting model could change when updating plugin version. Genrally speaking, there's no such thing like a `version` we could define to offer backward compatibility and load a yaml file defined for a previous version of Configuration-as-Code or for a distinct combination of jenkins-core + plugins.

This is simply false. Unlike the many other plugins for Jenkins, the CasC plugin does have control over it's own behavior. When the CasC plugin behavior changes in a way the changes the YAML schema, the CasC plugin can and should have a plan that minimize impact on users.

As I said in the discussion thread and in the other PR, this not a "Backward Compatibility" issue but a functional usability issue.

Re: [line 270](https://github.com/jenkinsci/jep/commit/adfe51ad87ac1870affbaf0d0ecbcc9f8418c41a#diff-d37718df6e1dbf6f51cbde624d32e737R270):

> So compatibility on version upgrade cannot be expected. But with JCasC we can produce a schema and tooling to discover such compatibility issue and let end-user know before they apply to production. Also, binary compatibility requirement in Jenkins ecosystem will significantly limit the risk for a broken configuration.

How does binary compatibility limit risk from changes to the UI (which is what CasC depends on)?

The planned tooling should to be included in the specification.

ndeloof:
> CasC plugin could generate a schema file for a particular configuration and inform the user when that schema changes in a breaking way.

CasC do already generate such a schema, but we just can't tell about schema changes : we only know about the current one. 

> This is simply false. Unlike the many other plugins for Jenkins, the CasC plugin does have control over it's own behavior.

From a user perspective, when you upgrade a plugin, being CasC or another one, model could change. So for sure we could write tons on `if version == 1 { } else { }` but this wouldn't change the general compatibility issue.

> How does binary compatibility limit risk ?

When some code define a UI property (getter/setter/constructor) it's part of the "public API" (!) and has to be kept. So plugin developer won't update this just by pleasure.

oleg-nenashev:
@bitwiseman Are you satisfied by the response?

P.S: I think the JEP document should be clarified according to the discussion before accepting it. CC @ewelinawilkosz

PR #108, opened 2018-05-15
re-accept JEP-201

ndeloof:


ndeloof:
@kohsuke please review

ewelinawilkosz:
to be completely honest I am a little bit lost - #106 was merged but @bitwiseman raised some questions, @ndeloof answered but had a question too... so I'd like JEP-201 to be re-approved but I feel we need to make sure we're more or less on the same page regarding #106 before doing that

ndeloof:
@kohsuke wording used in the Versioning section wasn't expected to dismiss this concern, but to make it clear this "expected feature" isn't under Configuration-as-Code scope as defined by this JEP. This doesn't mean a future, alternate implementation wouldn't provide some more advanced tooling or flexibility, just this specification is about the CasC mechanisms, not backward compatibility tooling. I'll submit another JEP on this topic, but I'd like to keep them separated so we don't have to amend JEP-201 for each and every concern about CasC usage. JEP-201 do define the mechanisms used to express Jenkins configuration as code with yaml files, detecting compatibility breaks with component versions is a higher level "usability" topic.

ndeloof:
see https://github.com/jenkinsci/jep/pull/111

bitwiseman:
I've had an offline discussion with @ewelinawilkosz and @ndeloof about this JEP.  I'll be submitting a PR for this JEP to get it into a state that it can be accepted as "Mission and Goals" JEP.

PR #110, opened 2018-05-21
Switch to a more understandable error status

rtyler:
I meant to mention this last week but forgot, "KO" as a status code is not obviously an error.  (looks just like a typo)


batmat:
:man_facepalming:

batmat:
BTW now I wonder if using `SUCCESS` for success would also a bit better than `OK`.

PR #111, opened 2018-05-30
Configuration as Code tooling proposal

ndeloof:
@ewelinawilkosz 
@bitwiseman 
@rtyler

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -26,7 +26,7 @@ endif::[]
 | Not Submitted :information_source:
 
 | Type
-| :bulb: Informational :bulb:
+| :bulb: Standards :bulb:
 
 | Created
 | :bulb: May 30th 2018 :bulb:
@@ -66,12 +66,12 @@ endif::[]
 
 == Abstract
 
-Configuration-as-Code (JEP-201) do describe the mechanisms used to configure a Jenkins master based
+Configuration-as-Code (JEP-201) describes the mechanisms used to configure a Jenkins master based
 on yaml file(s), by discovering the data model from Java API, without the need to write any "glue-code"
 for each and every supported plugin.
 
 This approach comes with some drawback: Any update to Jenkins master's core or plugins will change the
-inferred model, and as such a yaml configuration used a t time might not be valid a t+1. This concern
+inferred model, and as such a yaml configuration used a `t` time might not be valid at `t+1`. This concern
 has been raised by many early-adopters.
 
 This proposal is about defining the adequate tooling to balance this limitation, so both end-users
@@ -83,7 +83,7 @@ resolution hints.
 Configuration-as-Code tooling is designed to fill the gap between the raw mechanism defined by JEP-201
 to configure a specific jenkins-core + plugins combination, and the need to keep this configuration
 stable during core or plugin upgrades, or at least detect breaking changes and provide hints to manage
-upgreade.
+upgrade.
 
 This includes:
 
@@ -99,7 +99,7 @@ on this, when yaml validation do detect a configuration mismatch it can offer mi
 == Motivation
 
 Configuration-as-Code doesn't define a data model to setup jenkins, it's based on inferring a model
-from live jenkins instance from it's Java API, especially UI data-binding mechanisms. Based on this,
+from live jenkins instance from its Java API, especially UI data-binding mechanisms. Based on this,
 changes made to core or a plugin _might_ change this model and a configuration file _could_ not be
 valid after upgrade.
 
@@ -109,7 +109,7 @@ distinct models, as we only know about the live, inferred one.
 
 Configuration-as-Code tooling is defined to workaround this limitation by providing a toolset to
 discover such breaking changes early, before the actual production configuration is updated.
-It will help detect breaking changes during jenkins and plugins developement, offer an option for
+It will help detect breaking changes during jenkins and plugins development, offer an option for
 plugin developers to document required breaking changes, and validate a configuration for a specific
 jenkins instance before it get actually applied.
 
@@ -186,7 +186,7 @@ On end-user side we want ot offer a validation toolkit to be used when applying 
 instance. This toolkit do parse the yaml configuration, try to apply configuration on the inferred model,
 and will report issue detected during this dry-run configuration process.
 
-It would rely on breaking changes metadata to offer end-user some guidance on way to address repoerted
+It would rely on breaking changes metadata to offer end-user some guidance on way to address reported
 backward compatibility issues.
 
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,220 @@
+= JEP-0000: :bulb: Configuration as Code tooling :bulb:
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
+
+| Title
+| :bulb: Configuration as Code tooling :bulb:
+
+| Sponsor
+| :bulb: TBD :bulb:
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| :bulb: Informational :bulb:
+
+| Created
+| :bulb: May 30th 2018 :bulb:
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+//| JIRA
+//| :bulb: https://issues.jenkins-ci.org/browse/JENKINS-nnnnn[JENKINS-nnnnn] :bulb:
+//
+//
+// Uncomment if there will be a BDFL delegate for this JEP.
+//| BDFL-Delegate
+//| :bulb: Link to github user page :bulb:
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
+//
+//
+// Uncomment if this JEP depends on one or more other JEPs.
+| Requires
+| :bulb: JEP-201 :bulb:
+//
+//
+// Uncomment and fill if this JEP is rendered obsolete by a later JEP
+//| Superseded-By
+//| :bulb: JEP-NUMBER :bulb:
+//
+//
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+
+== Abstract
+
+Configuration-as-Code (JEP-201) do describe the mechanisms used to configure a Jenkins master based
+on yaml file(s), by discovering the data model from Java API, without the need to write any "glue-code"
+for each and every supported plugin.
+
+This approach comes with some drawback: Any update to Jenkins master's core or plugins will change the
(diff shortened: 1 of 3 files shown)


ewelinawilkosz:
issue raised some discussions already and we won't find a silver bullet, but if we have a dry run I can live with proposed solution 👍 
also once the plugin becomes mature and popular I assume we can hope plugins' maintainers may consider impact on JCasC  when implementing changes

bitwiseman:
@ndeloof Are you asking for commentary from me as a JEP editor or as a contributor?   

As an editor, I would say theres a bunch of stuff to be cleaned up (the `:bulb:` entries) and you need to flesh out the "Specification" section more.  Some of that would come from parts of your "Reasoning" section.   Description of the design that was chosen and is being implemented goes in the "Specification", discussion of alternate designs that were rejected and explanation as to why goes in the "Reasoning" - See the [section descriptions for "Specification" and "Reasoning" in JEP-1](https://github.com/jenkinsci/jep/tree/master/jep/1#required-sections).

ndeloof:
@bitwiseman your comment is actually useful for me to understand which information are required for such a JEP. I have to admit this highly standardized process is a new way for me to consider software development

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+changes they made during development might have impact on Configuration-as-Code.
+- Document recommendations for plugin developers on ways to avoid breaking changes
+- Metadata definition for plugin developers to document breaking changes and how to fix them. Based
+on this, when yaml validation do detect a configuration mismatch it can offer migration hints.
+
+
+== Motivation
+
+Configuration-as-Code doesn't define a data model to setup jenkins, it's based on inferring a model
+from live jenkins instance from it's Java API, especially UI data-binding mechanisms. Based on this,
```
jglick: `its`

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+changes made to core or a plugin _might_ change this model and a configuration file _could_ not be
+valid after upgrade.
+
+There's not so much cases where such compatibility breaks could happen, but there's chances this would.
+Compared to other tools, like `docker-compose`, we can't provide a `version` attribute and accept
+distinct models, as we only know about the live, inferred one.
+
+Configuration-as-Code tooling is defined to workaround this limitation by providing a toolset to
```
jglick: `work around`

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+=== Relax Configuration-asCode for backward compatibility ?
+
+There's few circumstances where we have options
+to workaround a technical change in Jenkins or plugin codebase as the technical bits are still available,
+just hidden by the default "model inference" mechanism.
+
+A typical sample for this has been debated on
+https://github.com/jenkinsci/configuration-as-code-plugin/issues/233[issue 233]: `JNLPLauncher` DataBound
+constructor has been updated in jenkins codebase to introduce new argument `workDirSettings`. This one
+is declared in API as `@Nonnull`, which means Configuration-as-Code will infer a required attribute. As
```
jglick: Another evaluation would be that this change to `JNLPLauncher` was wrong and should be fixed: if there was a reasonable default for `workDirSettings`, then it should have been in a `@DataBoundSetter` and thus optional. In fact there _is_ such a [reasonable default](https://github.com/jenkinsci/jenkins/blob/d4737b2b60438ce0e3cb7a8a6f26675f7c880914/core/src/main/java/jenkins/slaves/RemotingWorkDirSettings.java#L55) as seen also [here](https://github.com/jenkinsci/jenkins/blob/d4737b2b60438ce0e3cb7a8a6f26675f7c880914/core/src/main/resources/jenkins/slaves/RemotingWorkDirSettings/config.jelly#L28-L37). So this was just a bug in core that should be fixed, not something to work around. CC @oleg-nenashev

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+=== During Jenkins development
+
+Actually the exact same code change discussed in https://github.com/jenkinsci/configuration-as-code-plugin/issues/233[issue 233]
+could have been implemented without breaking Configuration-as-Code inferred model :
+
+- by defining a `@DataBoundSetter` for newly introduced `workDirSettings` attribute
+- by not defining `workDirSettings` argument as `@Nonnull` and applying default value
+
+Assuming Configuration-as-Code would exist for some time when this change was applied to jenkins-core, we
+might have detected it during code review and prevented this to happen.
```
jglick: In fact this change to `JNLPLauncher` was wrong and should be fixed: if there was a reasonable default for `workDirSettings`, then it should have been in a `@DataBoundSetter` and thus optional. And there _is_ such a [reasonable default](https://github.com/jenkinsci/jenkins/blob/d4737b2b60438ce0e3cb7a8a6f26675f7c880914/core/src/main/java/jenkins/slaves/RemotingWorkDirSettings.java#L55) as seen also [here](https://github.com/jenkinsci/jenkins/blob/d4737b2b60438ce0e3cb7a8a6f26675f7c880914/core/src/main/resources/jenkins/slaves/RemotingWorkDirSettings/config.jelly#L28-L37). CC @oleg-nenashev

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+changes made to codebase so Configuration-as-Code can give end-user some hints on the break and how to fix
+it.
+
+=== Before applying configuration
+
+On end-user side we want ot offer a validation toolkit to be used when applying configuration on a jenkins
+instance. This toolkit do parse the yaml configuration, try to apply configuration on the inferred model,
+and will report issue detected during this dry-run configuration process.
+
+It would rely on breaking changes metadata to offer end-user some guidance on way to address repoerted
```
jglick: typo

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+will warn developer about breaking change being introduced for Configuration-as-Code users.
+
+Sometime such changes are actually required. For such situation we propose some metadata being added aside
+changes made to codebase so Configuration-as-Code can give end-user some hints on the break and how to fix
+it.
+
+=== Before applying configuration
+
+On end-user side we want ot offer a validation toolkit to be used when applying configuration on a jenkins
+instance. This toolkit do parse the yaml configuration, try to apply configuration on the inferred model,
```
jglick: BTW please follow the formatting model of one physical (AsciiDoc) text line per sentence or major phrase. Easier to read diffs and so on.

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+N/A
+
+== Security
+
+N/A
+
+== Infrastructure Requirements
+
+If approved, Configuration-as-Code tooling to check for API backward compatibility will need to be
+included in the CI process, so it can inform plugin developers and contributors on breaking changes
```
jglick: Meaning what, exactly? That an incompatible change to databinding produces a build failure? Even assuming the other pieces have been developed, this is not trivial as you need to determine a baseline to compare against, which presumably means looking up the most recent Maven release older than the current snapshot.

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+Compared to other tools, like `docker-compose`, we can't provide a `version` attribute and accept
+distinct models, as we only know about the live, inferred one.
+
+Configuration-as-Code tooling is defined to workaround this limitation by providing a toolset to
+discover such breaking changes early, before the actual production configuration is updated.
+It will help detect breaking changes during jenkins and plugins developement, offer an option for
+plugin developers to document required breaking changes, and validate a configuration for a specific
+jenkins instance before it get actually applied.
+
+== Reasoning
```
jglick: Something not addressed in [JEP-201 itself](https://github.com/jenkinsci/jep/blob/master/jep/201/README.adoc#general-mechanism) is why you chose to use databinding (~ `DescribableModel`) as the source of truth, when plugins have traditionally maintained settings compatibility via their XStream form (~ fields, `readResolve`)—we _do_ presume already that a `config.xml` which worked in an older “version” of Jenkins can be loaded safely in a newer version. Thus JEP-201 is introducing an _additional_ compatibility constraint: plugin developers need to be aware of, and work around, both models when refactoring settings.

(Pipeline was already imposing a similar constraint for `Step`s and those `Describable`s being used inside `Step`s. OTOH it also allows `StepDescriptor.newInstance` / `.uninstantiate` to be overridden to customize the process, as I in fact did recently in https://github.com/jenkinsci/docker-workflow-plugin/pull/138. This is not available when sharing code with freestyle via `CoreStep` / `CoreWrapperStep`. I think `job-dsl` also introduced the ability to introspect stuff and pay attention to `@Symbol` but AFAIK most bindings are still hand-written.)

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+changes made to core or a plugin _might_ change this model and a configuration file _could_ not be
+valid after upgrade.
+
+There's not so much cases where such compatibility breaks could happen, but there's chances this would.
+Compared to other tools, like `docker-compose`, we can't provide a `version` attribute and accept
+distinct models, as we only know about the live, inferred one.
+
+Configuration-as-Code tooling is defined to workaround this limitation by providing a toolset to
+discover such breaking changes early, before the actual production configuration is updated.
+It will help detect breaking changes during jenkins and plugins developement, offer an option for
```
jglick: typo

ndeloof:
@jglick I proposed another JEP following Jenkins Essentials model to have smaller JEPs focussed on some aspect of a general topic

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+
+| Sponsor
+| :bulb: TBD :bulb:
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| :bulb: Informational :bulb:
```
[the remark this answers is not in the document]
ndeloof: None of the proposed types do actually match what I try to express in this JEP. Sounds "_Standards_" is what is used for this purpose, but I'm not defining a standard here, just exploring the concrete tools that would help JEP-201 adoption.

jep/0000/README.adoc
jglick: You need not necessary “store an inferred model between builds”. You can have a build compute the old model almost as easily as it computes the new model. But you need some way to deterministically identify the “old version”.

jep/0000/README.adoc
```diff
@@ -0,0 +1,220 @@
...
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+
+== Abstract
+
+Configuration-as-Code (JEP-201) describes the mechanisms used to configure a Jenkins master based
+on yaml file(s), by discovering the data model from Java API, without the need to write any "glue-code"
```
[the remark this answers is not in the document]
ndeloof: Having a "Configurable" API implemented by all plugin would indeed ensure the API is well handled by plugin maintainer and that it correctly reflects the configurable elements of a plugin,  but doing so would require years before we get good coverage from all major plugins (just see how long it took for pipeline adoption). 

By the way, this "no glue code" approach, based on UI databing exposed model, is not as bad as you can expect, and easy for plugin maintainer to improve as it relies on the databinding mechanisms they're used to.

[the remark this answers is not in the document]
bitwiseman: @amuniz 
What you suggest is not in opposition to avoiding glue-code.  The details of this should be discussed in this upcoming JEP submission for Jenkins CasC Schema.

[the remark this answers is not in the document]
amuniz: > upcoming JEP submission for Jenkins CasC Schema

Sorry, I might be a bit lost... is not the schema already defined? It is using the existing model with `@DataBoundContructor`/`@DataBoundSetter` introspection, isn't it?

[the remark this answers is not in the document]
ndeloof: This will be presented in more details (reasoning, wroakrounds, etc) in https://github.com/jenkinsci/jep/pull/125

PR #118, opened 2018-06-06
Add JEP for ruby-runtime deprecation

daniel-beck:
[Previously](https://groups.google.com/d/msg/jenkinsci-dev/Ve0fqAud3Mk/MTIxw6ZyBwAJ)

### DIFF
--- a/jep-0000/README.adoc
+++ b/jep-0000/README.adoc
@@ -1,0 +1,153 @@
+= JEP-0000: Deprecation of ruby-runtime
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
+
+| Title
+| Deprecation of ruby-runtime
+
+| Sponsor
+| https://github.com/daniel-beck
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| Process
+
+| Created
+| 2018-06-06
+
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+//| JIRA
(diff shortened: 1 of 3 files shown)


jep-0000/README.adoc
```diff
@@ -0,0 +1,153 @@
...
+//
+//
+// Uncomment if there will be a BDFL delegate for this JEP.
+//| BDFL-Delegate
+//| :bulb: Link to github user page :bulb:
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
```
oleg-nenashev: Link the pending thread here

jep-0000/README.adoc
```diff
@@ -0,0 +1,153 @@
...
+== Security
+
+There are no security risks related to this proposal.
+
+
+== Infrastructure Requirements
+
+This JEP will be implemented by using a well established feature of the update center generator.
+
+There are no new infrastructure requirements related to this proposal.
```
oleg-nenashev: Unless we implement the new-update site proposal in the thread

oleg-nenashev: it would be also great to still list these plugins in plugins.jenkins.io (with a "deprecated" warning and link to the download guidelines)

jep-0000/README.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+
+== Testing
+
+There are no testing issues related to this proposal.
```
oleg-nenashev: We need to ensure that there is no tests using the plugins in ATH, and so on

bitwiseman:
Okay, I'm assigning JEP-7 to this submission.

jep-0000/README.adoc
daniel-beck: It absolutely would be, but it's not possible with what we have.

BobDu:
I am confused to see that the JEP status is Draft.
Is it because it lacks updates after its implementation? Or is there indeed something that has not been done?

@daniel-beck  Sorry to disturb you.

daniel-beck:
@BobDu We're generally not doing great at keeping JEPs synched with reality (to clarify, not blaming anyone). This one has been implemented (note the substantial amendments in https://github.com/jenkinsci/jep/pull/356).

PR #125, opened 2018-06-16
Configuration as Code YAML Schema

bitwiseman:
### NOT READY FOR DRAFT

@ndeloof @ewelinawilkosz 
I pulled the text from JEP-201 to here.  This is a placeholder for the Configuration as Code YAML Schema JEP.  

Before this can be approved as Draft,  it seems to me that most of the Specification needs rewriting/filling in - describing how the schema we be generated and the APIs for customizing the schema when the automatically generated schema is not sufficient. 

However, I don't have the expertise to write that. 

You can either update this PR or pull the branch and start your own.

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -164,7 +164,7 @@ We need some mechanism to construct a jenkins component in respect with it's lif
 schema is relevant for end-user based on his use of jenkins and it's web UI. Such component are tightly coupled with
 Stapler web framework and the expectation for them to be managed by this framework.
 
-++DataBoundSetter++s & ++DataBoundConstructor++s offer a natural way to construct Jenkins components from a set of key=value pairs. Most
+``DataBoundSetter``s & ``DataBoundConstructor``s offer a natural way to construct Jenkins components from a set of key=value pairs. Most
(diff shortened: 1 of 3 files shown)


bitwiseman:
@ndeloof @ewelinawilkosz This needs to be updated.  Please make the edits and submit a PR.

ndeloof:
@bitwiseman yes I know, I will look into this JEP effort next week

ewelinawilkosz:
I will have a look today and update where it makes sense, thanks @bitwiseman!

jep/201/README.adoc
```diff
@@ -1,296 +0,0 @@
-= JEP-201: Jenkins Configuration as Code
```
oleg-nenashev: hmm, why does it delete the JEP entirely?

ndeloof: I guess @bitwiseman did moved jep-201 REAME file into this new template file, not just the content, so from a git history pov the initial JEP has been removed.

ndeloof: force-pushed to fix initial commit

bitwiseman:
Approving as JEP-209.  I think this needs more work and discussion, but it is ready for draft.

PR #128, opened 2018-06-22
Use UTF-8 for Pipeline build logs

jglick:
Submitted for approval as draft JEP by @jenkinsci/jep-editors.

@kohsuke I do not know who the BDFL delegate should be.

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,246 @@
+= JEP-0000: Use UTF-8 for Pipeline build logs
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
+
+| Title
+| Use UTF-8 for Pipeline build logs
+
+| Sponsor
+| link:https://github.com/jglick[Jesse Glick]
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| Standards
+
+| Created
+| :bulb: Date (YYYY-MM-DD) :bulb:
+
+| BDFL-Delegate
+| :bulb: Link to github user page :bulb:
+
+| JIRA
+| link:https://issues.jenkins-ci.org/browse/JENKINS-31096[JENKINS-31096]
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
+//
+//
+// Uncomment if this JEP depends on one or more other JEPs.
+//| Requires
+//| :bulb: JEP-NUMBER, JEP-NUMBER... :bulb:
+//
+//
+// Uncomment and fill if this JEP is rendered obsolete by a later JEP
+//| Superseded-By
+//| :bulb: JEP-NUMBER :bulb:
+//
+//
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+All Jenkins Pipeline build logs are stored internally in UTF-8 encoding.
+(The log can be seen by users from the *Console* link,
+but also from Blue Ocean, *Pipeline Steps*, and REST and CLI.)
+
+Text produced by Jenkins (Java) code is intrinsically Unicode.
+Output from external processes which was not already in UTF-8 is converted to UTF-8 as it is produced.
+
+The scope does not include other project types (freestyle),
+or other types of logging (system, branch indexing).
+
+== Specification
+
+Newly created Pipeline build logs use UTF-8 encoding for all text.
+This is hard-coded in the `charset` field in `build.xml`.
+The `TaskListener` associated with the build as a whole (`FlowExecutionOwner.getListener`),
+and with individual steps (`StepContext.get(TaskListener.class)`),
+will thus be using UTF-8 encoding.
+
+When output is produced by `DurableTaskStep` (`sh`, `bat`, `powershell`),
+it is transcoded to UTF-8 where necessary by `FileMonitoringController`.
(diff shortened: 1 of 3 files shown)


jep/0000/README.adoc
```diff
@@ -0,0 +1,246 @@
...
+is copied directly to the build log as a bytestream.
+Content rendered over HTTP is always sent in UTF-8 encoding.
+
+When Pipeline was introduced,
+this system could not applied as is
+since a Pipeline build could use many nodes, or none.
+The build log was just left in the _master_’s system default encoding;
+if that encoding happens to be identical to that used by an agent,
+then process output will be encoded directly,
+otherwise you get mojibake.
```
svanoort: Probably worth using a different term than "mojibake" because not all JEPs will be read by people familiar with this class of encoding issue.  I suggest 'garbled text for invalid characters ("mojibake")' to clarify.

jep/0000/README.adoc
```diff
@@ -0,0 +1,246 @@
...
+should see no change in behavior.
+
+=== Historical builds
+
+Historical builds may have recorded a different `charset` in `build.xml`.
+In such a case, their log text will continue to be served in that encoding.
+
+If the build was started before the upgrade but is still running,
+it will continue to use the recorded encoding.
+That may mean that newly produced text contains mojibake.
```
svanoort: I'd like to know if there's a way we could avoid that with special-casing for builds that were in-flight at time of upgrade?

jep/0000/README.adoc
```diff
@@ -0,0 +1,246 @@
...
+
+=== Partial updates
+
+If the Jenkins administrator updates one of `workflow-job` or `workflow-durable-task-step`,
```
svanoort: This one doesn't worry me so much.

jep/0000/README.adoc
```diff
@@ -0,0 +1,246 @@
...
+to users accessing the `$JENKINS_HOME/jobs/…/builds/…/log` file directly.
+While that may have been common in the early days of Hudson,
+this is likely a rare use case today.
+
+In any event, making Jenkins architecture more complex and potentially buggy
+to satisfy a marginal requirement seems a poor trade-off.
+
+== Backwards Compatibility
+
+Jenkins systems running on UTF-8-based computers
```
svanoort: And Jenkins systems running without UTF-8?  The *very precise* wording here makes me nervous because it's emphasizing that it'll work in one particularly common case but is not making other promises beyond that.

jglick:
> The majority of the issues I've seen are caused by agents whose default encoding is ANSI_X3.4-1968 (US-ASCII)

Which sorts of issues? Or you mean _masters_ set to ASCII, which would definitely cause problems?

jglick:
> [I…] am not sure this is something that should actually _require_ a JEP.

Perhaps not, but I needed to write down some of this reasoning anyway, and this was as easy as anything else.

For context, there is a bigger set of changes linked to JENKINS-38381 (JEP(s) to be filed) of which this is a prerequisite, so I wanted a clear trail of the dependencies.

jep/0000/README.adoc
jglick: That is quite a mouthful to repeat. I will try to explain and hyperlink on first usage.

jep/0000/README.adoc
```diff
@@ -0,0 +1,246 @@
...
+When Pipeline was introduced,
+this system could not applied as is
+since a Pipeline build could use many nodes, or none.
+The build log was just left in the _master_’s system default encoding;
+if that encoding happens to be identical to that used by an agent,
+then process output will be encoded directly,
+otherwise you get mojibake.
+
+Pipeline also suffered from a problem that continues to affect freestyle builds:
+non-ASCII text send directly by Java code into the build log
```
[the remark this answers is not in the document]
jglick: oops, will fix

jep/0000/README.adoc
jglick: I am just trying to start the compatibility section with the broad point that if you set UTF-8 as your system encoding everywhere, you can ignore the subsections. Will try to clarify.

jep/0000/README.adoc
jglick: Possibly, but it is hardly worth the bother for a one-time event, when the great majority of logs contain no non-ASCII characters, and any impact even on this one build is minimal.

dwnusbaum:
> Which sorts of issues? Or you mean masters set to ASCII, which would definitely cause problems?

Usually text coming directly from Java code such as your `Folder Name » Job Name` example, which gets turned into `Folder Name ? Job Name`.

jglick:
> text coming directly from Java code

Right, so this has nothing to do with the system encoding on agent computers. It is just an effect of the historical `WorkflowRun.charset` coming from the master’s `Charset.defaultCharset` and potentially being a non-Unicode encoding incapable of handling the `»` character from the Latin-1 Supplement block. That is fixed by this JEP.

rtyler:
I think once we can figure out a suitable BDFL Delegate to throw into this Draft, then we can merge it.

jep/0000/README.adoc
svanoort: Users seem to get disproportionately angry when an upgrade introduces issues, even if it's just for the current build  -- but I'll buy that this will be fairly rare outside Japan, Russia, China, and South Korea.   So, what would it take to prevent impacting those countries?

jep/0000/README.adoc
```diff
@@ -0,0 +1,256 @@
...
+
+When Pipeline was introduced,
+this system could not applied as is
+since a Pipeline build could use many nodes, or none.
+The build log was just left in the _master_’s system default encoding;
+if that encoding happens to be identical to that used by an agent,
+then process output will be encoded directly.
+Otherwise you get
+link:https://en.wikipedia.org/wiki/Mojibake[mojibake]:
+garbled text caused by bytes being interpreted as the wrong characters.
```
svanoort: Much better for the audience, thanks!

jep/0000/README.adoc
svanoort: Note: happy to pick that thread up in the PRs -- IIRC this was my initial concern and one of the main reasons it didn't get fast-merged ASAP.

jep/0000/README.adoc
```diff
@@ -0,0 +1,256 @@
...
+| Not Submitted :information_source:
+
+| Type
+| Standards
+
+| Created
+| 2018-06-22
+
+| BDFL-Delegate
+| :bulb: Link to github user page :bulb:
```
bitwiseman: Who should be the BDFL Delegate for this?

svanoort: @bitwiseman I'm willing to take it if KK approves that -- and have already given my approval for the JEP so if he approves my approval, then it's merge-able.

PR #132, opened 2018-06-28
Accept JEP-201

bitwiseman:
@oleg-nenashev @jglick @kohsuke 
https://github.com/jenkinsci/jep/pull/108 provides context for the changes made to this JEP. 
There are at least to follow up JEPs in the pipe: tooling and schema.

### DIFF
--- a/jep/201/README.adoc
+++ b/jep/201/README.adoc
@@ -22,7 +22,7 @@ endif::[]
 | https://github.com/ewelinawilkosz2[Ewelina Wilkosz] from https://github.com/praqma[Praqma]
 
 | Status
-| Draft :speech_balloon:
+| Accepted :ok_hand:
 //| Rejected :no_entry:
 //| Withdrawn :hand:
 //| Final :lock:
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Needs BDFL Delegate approval before merging. It's not clear to me whether there is such approval in #108

bitwiseman:
@oleg-nenashev discussing with KK now.

daniel-beck:
Didn't we have feedback from @jglick critical about the "natural" structure from XML hierarchy? If so, are we no longer including substantial feedback in some "answered but rejected" form in JEPs?

daniel-beck:
(Please note I'm not a JEP editor, so my review means nothing. Commit access via org ownership.)

FWIW the section 'Intuitive schema' is included here but no concern about the status of Blue Ocean is mentioned, but if @bitwiseman things it can be deferred, I'm not opposed.

bitwiseman:
I discussed this with @kohsuke and he verbally said he agreed with this being accepted.  He's traveling  now. I'll go ahead and merge this change.

PR #135, opened 2018-07-04
Platform SiG JEP Draft

oleg-nenashev:
As a follow-up for the Jenkins & Java 10+ Hackathon discussions, I would like to propose a Platform SiG as a venue for discussions.

Motivation: Jenkins is being used in various configurations and environments. Although Jenkins is mainly implemented in Java, it is not just “write once, run everywhere”. It is a complex solution, which includes special packaging and native code in some cases (e.g. Windows Service Wrapper, Windows Process Management Lib, LibZFS, etc.). The purpose of this SiG is to focus on such platforms and Packaging related to Jenkins core.

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,165 @@
+= JEP-0000: Platform Special Interest Group
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
+
+| Title
+| Platform Special Interest Group (Platform SiG)
+
+| Sponsor
+| link:https://github.com/oleg-nenashev[Oleg Nenashev]
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| Informational
+
+| Created
+| 2018-07-04
+
+| BDFL-Delegate
+| :bulb: Link to github user page :bulb:
+
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+//| JIRA
+//| :bulb: https://issues.jenkins-ci.org/browse/JENKINS-nnnnn[JENKINS-nnnnn] :bulb:
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
+//
+//
+// Uncomment if this JEP depends on one or more other JEPs.
+//| Requires
+//| :bulb: JEP-NUMBER, JEP-NUMBER... :bulb:
+//
+//
+// Uncomment and fill if this JEP is rendered obsolete by a later JEP
+//| Superseded-By
+//| :bulb: JEP-NUMBER :bulb:
+//
+//
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+## Abstract
+
+This informational Jenkins Enhancement Proposal describes the Platform Special Interest Group,
+which has been proposed as a part of the Jenkins & Java 10+ Hackathon.
+The Group offers a place for all kinds of platform support discussions:
+Java, Operating Systems, Architectures, Docker, Packaging, Web Containers, etc.
+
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
CC @bitwiseman and @tracymiranda

jep/0000/README.adoc
```diff
@@ -0,0 +1,161 @@
...
+| Created
+| 2018-07-04
+
+| BDFL-Delegate
+| :bulb: Link to github user page :bulb:
+
+| JIRA
+| link:https://issues.jenkins-ci.org/browse/INFRA-1701[INFRA-1701]
+
+| Discussions-To
```
bitwiseman: This is only needed if discussion will occur on a ML other than jenkinsci-dev.

tracymiranda:
The JEP process is fairly new so apologies as we work to make it more effective. Couple of things to note (@bitwiseman please correct as necessary):

- no BDFL delegate means delegate defaults to Kohsuke
- draft state is not meant to be a blocker. Purpose of JEP is to share proposals in transparent way, but we can progress thing before they move into an accepted state.
- the process as you note is flawed because there are no ETAs on state of JEPs, @bitwiseman and I are working on that, current thinking is a regular bi-weekly JEP update session

So regarding this JEP & JEP-4 (now that I have read all & understand things better)
- I agree with @bitwiseman that this JEP is unnecessary and we should use JEP-4 as the umbrella for SIGs
- That being said, we need a space for the info in this doc, particularly the charter, SIG lead & participants to be shared, ideally on jenkins.io. We need to figure this out as next step soon.
- We have a temporary bottleneck as our BDFLs take a break, so might be a couple of weeks before we get all plumbing in place. Hopefully no more than that. 
- In the meantime, launching it on jenkins-dev is good. I suggest next steps are to progress on drumming up interest and socialize with the people who will eventually participate. Perhaps a target date for first live meetup will be good. 
- Thanks for being the first SiG to put through the process!

oleg-nenashev:
@tracymiranda 

> I agree with @bitwiseman that this JEP is unnecessary and we should use JEP-4 as the umbrella for SIGs

Once JEP-4 is landed, I agree. But for now I would rather prefer to have a separate JEP as long as it is consistent with JEP-4. It makes the proposal more legitimate in terms of the process, even if we start SiG before the JEP is fully approved

> That being said, we need a space for the info in this doc, particularly the charter, SIG lead & participants to be shared, ideally on jenkins.io. We need to figure this out as next step soon.

This is something I can help with if it helps to deliver Platform SiG

> I suggest next steps are to progress on drumming up interest and socialize with the people who will eventually participate. Perhaps a target date for first live meetup will be goo

Yes, I could do that. And #137 is the topic which we can discuss at the first meeting

bitwiseman:
@oleg-nenashev So, this PR isn't needed anymore right?  The work on JEP-4 is proceeding and the SIG has been created based on that work.

oleg-nenashev:
@bitwiseman please keep it open for now. I will repurpose it to define special requirements for the SIG as I proposed in #138

jep/0000/README.adoc
```diff
@@ -0,0 +1,161 @@
...
+Meetings will be conducted and recorded via Jenkins Hangouts-on-Air.
+
+### Chairs
+
+* Oleg Nenashev
+* TBD
+
+### Channels
+
+* `jenkins-platform-sig` Google group for internal sync-ups
```
MarkEWaite: Since the Google group can be represented as a URL, could this be changed to

> * `link:https://groups.google.com/forum/#!forum/jenkins-platform-sig[jenkins-platform-sig]` Google group for internal sync-ups

jep/0000/README.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+This section will be updated according to the feedback in developer mailing lists.
+
+## Backwards Compatibility
+N/A - Informational JEP
+
+## Security
+
+* No special Security team requirements for the implementation
+* At least one of Platform SiG
```
MarkEWaite: I don't understand what is meant by

> * At least one of Platform SiG

with regards to security.  Is it that the Platform SIG would like at least one member of the security team to also be part of the SIG, or is it that the Platform SIG has some security-related actions that they do at least once?

jep/0000/README.adoc
```diff
@@ -0,0 +1,161 @@
...
+* SiG may be involved into reviews of security issues related to specific platforms
+
+## Infrastructure Requirements
+
+The group will be using the existing Jenkins infrastructure: Jenkins Developer mailing List, etc. New Google groups and channels will be created on-demand. There are the following new requirements:
+
+* New Google group is established: “jenkins-platform-sig”
```
MarkEWaite: Hyperlink to the Google group?

> * New Google group is established: “link:https://groups.google.com/forum/#!forum/jenkins-platform-sig[jenkins-platform-sig]”

bitwiseman:
@oleg-nenashev 
It is almost two months since last update.  
PR for work in progress makes sense only if the work is progressing. 😄 
Closing this PR doesn't loose the branch or work done on it.

oleg-nenashev:
Do not care much anymore. The SIG is up and running well

PR #136, opened 2018-07-04
Cloud Native SIG JEP draft

carlossg:
The Cloud Native group of contributors and collaborators focuses on improving Jenkins to run on Cloud environments as a "Cloud Native" application.

cc @tracymiranda


oleg-nenashev:
Related proposal: #135 . We do not have a standard template for SiG JEPs, so the contents are different a bit

jglick:
FTR my suggestion on the dev list was that this be reopened as some kind of umbrella JEP to give broader context for JEP-202 and upcoming efforts.

oleg-nenashev:
I have proposed edits for JEP&SIG in https://github.com/jenkinsci/jep/pull/138, waiting for feedback from reviewers

bitwiseman:
@jglick The SIG is still going to happen it just won't need a JEP.

PR #138, opened 2018-07-08
[JEP-4] - Change proposals after creating Platform SIG

oleg-nenashev:
Last week we have created the first SIG in Jenkins by following the process in this SIG. We got https://jenkins.io/sigs/ and https://jenkins.io/sigs/platform/ hosted.

This pull request reflects some of my experience:

- [x] - Added links to the reference implementation
- [x] - Added "SIG id" to unify channel naming
- [x] - Added SIG metadata to the JEP, adjusted naming conventions to follow it
- [x] - Made GitHub aliases optional
- [x] - Adjusted the minimal frequency of meetings to "once per month"
- [x] - Added "Chats" section
- [x] - Added "Transferring leadership and adopting SIGs" section

CC @rtyler @bitwiseman @tracymiranda

### DIFF
--- a/jep/4/README.adoc
+++ b/jep/4/README.adoc
@@ -44,7 +44,7 @@ endif::[]
 //
 // Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
 | Discussions-To
-|
+| link:https://groups.google.com/d/topic/jenkinsci-dev/6-1mZoKp4hM/discussion[jenkinsci-dev@ thread]
 //
 //
 // Uncomment if this JEP depends on one or more other JEPs.
@@ -83,7 +83,7 @@ In order to standardize Special Interest Group efforts, create maximum
 transparency, and route contributors to the appropriate SIG, SIGs should follow
 the guidelines stated below:
 
-* Meet regularly, at least for 30 minutes every 3 weeks, except November and December
+* Meet regularly, at least for 30 minutes every month
 * Keep up-to-date meeting notes, somewhere publicly linked from the SIG's page on jenkins.io
 * Announce meeting agenda and minutes after each meeting, on their SIG mailing list
 * Record SIG meeting, either via <<video, YouTube Live Streaming>>, or via
@@ -92,8 +92,7 @@ the guidelines stated below:
   (video or text) or the meeting minutes published afterwards.
 * Ensure the SIG's mailing list, and optional chat channel are archived
 * Report activity in the link:https://wiki.jenkins.io/display/JENKINS/Governance+Meeting+Agenda?cache[bi-weekly Governance Meeting] at least once every 6 weeks
-* Participate in release planning meetings and retrospectives, and burndown
-  meetings, as needed.
+* Participate in Governance meetings, as needed.
 * Ensure related work happens in a project-owned GitHub org and repository,
   with code and tests explicitly owned and supported by the SIG, including issue
   triage, PR reviews, test-failure response, bug fixes, etc.
@@ -108,7 +107,7 @@ link:https://jenkins.io/conduct[Code of Conduct].
 
 - **SIG Participant**: active in one or more areas of the project; wide
   variety of roles are represented
-- **SIG Lead**: SIG organizer
+- **SIG Lead**: SIG organizer. SIG may have multiple leaders
 
 === Creation and Maintenance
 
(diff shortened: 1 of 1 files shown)


jep/4/README.adoc
```diff
@@ -44,7 +44,7 @@ endif::[]
 //
 // Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
 | Discussions-To
-|
+| link:https://groups.google.com/d/topic/jenkinsci-dev/6-1mZoKp4hM/discussion[jenkinsci-dev@ thread]
```
bitwiseman: @oleg-nenashev This field is only needed if the discussions in general will occur on a mailing list other than `jenkinsci-dev`

oleg-nenashev:
@bitwiseman adding the link was intentional. For JEP visitors it may be helpful to see a link to the discussion thread without looking for it in the mailing list.

bitwiseman:
@oleg-nenashev 
Understood but that is not the described purpose of that field.  If you'd like to change that, please start discussion and/or PR against JEP-1 to add such.

jep/4/README.adoc
```diff
@@ -147,21 +163,21 @@ Create Google Groups at
...
   via email address (with a suitable welcome message); in Members/All Members
-  select Tyler and Olivier and assign t hem to an "owner role" for long term
+  select Tyler and Olivier and assign them to an "owner role" for long term
   maintenance.
 * Set "View topics", "Post", "Join the Group" permissions to be "Public"
 
-Name convention:
-
-* jenkins-foo (the discussion group)
+Naming convention: `jenkins-${sigId}` (the discussion group)
```
bitwiseman: The existing groups are name `jenkinsci-<etc>`.   Is the difference here intentional?

oleg-nenashev: I followed the original naming format from @rtyler . Generally jenkins vs. jenkinsci is a long-long story. I have no preference., both fine with me

jep/4/README.adoc
```diff
@@ -194,24 +210,45 @@ and should consult the link:https://jenkins.io/event-calendar/[Jenkins Event
 Calendar] to make sure their meetings do not overlap with those already
 scheduled.
 
+[[chat]]
+==== Chats
+
+SIGs can *optionally* create dedicated channels for chats.
+These chats may be located in IRC, Gitter or other channels.
```
bitwiseman: Perhaps we could maintain a list of platforms that are acceptable.   We don't want official SIG chats hosted on obscure or add platforms.

casz:
Perhaps another medium instead of mailing list would be GitHub's team discussion? https://blog.github.com/2017-11-20-introducing-team-discussions/

jep/4/README.adoc
```diff
@@ -147,21 +163,21 @@ Create Google Groups at
 link:https://groups.google.com/forum/#!creategroup[],
 following the procedure:
 
-* Each SIG should have one discussion groups, and a number of groups for mirroring relevant github notifications;
+* Each SIG should have at least one discussion group.
```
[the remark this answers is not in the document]
oleg-nenashev: Yeah should/must discussion from JEP-1 :) Will fix it

oleg-nenashev:
@casz Looks interesting, I am open to try it. We are not happy about Google Groups in some cases (like Jenkinsci DEV being blocked sometimes), and a GitHub-binded channel could be a good option. Started the test discussion in https://github.com/orgs/jenkinsci/teams/java10-support/discussions/1

oleg-nenashev:
Unfortunately Discussions have no auto-linking like pull requests and issues do. But it looks like an interesting option to consider

casz:
If you use some extensions for github it can look nice :)
image
I use refined GitHub https://github.com/sindresorhus/refined-github and https://github.com/StylishThemes/GitHub-Dark which to me makes my GitHub experience like I want it 👍

jep/4/README.adoc
bitwiseman: So far we've used `jenkins-${sigId}-sig`.  Which seems fine to me.

oleg-nenashev:
@bitwiseman before it is merged, I believe it should be approved by the JEP sponsor. If the JEP sponsor is not available (it's what I see from the ML), the JEP should be handed over to another sponsor before merging

oleg-nenashev:
BTW, the organization wording looks good to me. I would be explicit that orgs are listed just for visualization, and that all active participants should be still listed as participants. It means that we do not add an entire organization as a group eligible for voting, for example

PR #140, opened 2018-07-10
[JENKINS-52210] Essentials AWS auto-configuration

batmat:
[JENKINS-52210](https://issues.jenkins-ci.org/browse/JENKINS-52210)

cc @mandie722 @rtyler

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -123,11 +123,9 @@ A dedicated instance profile is created by the _CloudFormation_ template to allo
 
 ==== S3 bucket policy
 
-FIXME: refine through JENKINS-52342.
-
 [source,json]
 "Statement": [{
-        "Sid": "TodoRefineSecurityALot",
+        "Sid": "TodoRefineSecurity-JENKINS-52342",
         "Effect": "Allow",
         "Action": [
             "s3:PutObject",
@@ -138,6 +136,10 @@ FIXME: refine through JENKINS-52342.
         "Resource": "*"
     }
 ]
+
+CAUTION: Currently the prototype is configured with the way shown above.
+We would like to reduce the access to only the dedicated bucket created for Jenkins Essentials usage.
+We are tracking this improvement as link:https://issues.jenkins-ci.org/browse/JENKINS-52342[JENKINS-52342].
 
 === Security groups
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -150,6 +150,7 @@ A dedicated group is created to allow access to the following ports:
 The user will be strongly advised to use their own public IP to restrict SSH access to the master to her only (i.e. discourage `0.0.0.0/0`).
 footnote:[using `curl ident.me` is often to help find out one' public IP and pass the value to the CloudFormation template parameter.]
 8080:: not restricted.
+50000:: not restricted (for agents).
 
 ==== Agents
 
@@ -208,6 +209,10 @@ But SSH is deemed sufficiently secure so that we still allow users to pass `0.0.
 === HTTPS by default
 
 (?) TBD FIXME
+
+=== Agent port
+
+WIP: FIXME: do we even need to open it. Does the agent actually connect through the SSH pipe?
 
 == Infrastructure Requirements
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -186,9 +186,8 @@ Two more need to be re-enabled for the Essentials AWS flavor to be able to auto-
 
 A CloudFormation template parameter `SSHLocation` is used to define the IPs allowed to access the master using SSH.
 
-Users are strongly advised to pass only their own IP when creating their setup.
-
-// Should we actually fail if one passes 0.0.0.0/0?
+Users are advised to pass only their own IP when creating their setup.
+But SSH is deemed sufficiently secure so that we still allow users to pass `0.0.0.0/0` as a value if they want this.
 
 === HTTPS by default
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -206,9 +206,7 @@ If we want to be able to run automated tests, we need some form of AWS account t
 [[aws-testing]]
 === AWS testing of CloudFormation template
 
-Given an <<aws-account>>, we can use the `aws` CLI, we need to set up automated tests using the .
-
-It is straightforward to implement:
+Given an <<aws-account>>, we can use the `aws` CLI to automatically trigger the provisioning, then retrieve what is neeeded to run automated tests:
 
 * create the stack using `aws cloudformation create-stack ...`
 * retrieve the EC2 instance IP using:
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,232 @@
+= JEP-0000: Essentials AWS auto-configuration
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
+
+| Title
+| Essentials AWS auto-configuration
+
+| Sponsor
+| https://github.com/batmat[Baptiste Mathus]
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| Standards
+
+| Created
+| 2018-07-10
+
+| BDFL-Delegate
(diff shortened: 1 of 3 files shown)


jep/0000/README.adoc
```diff
@@ -0,0 +1,232 @@
...
+
+=== Security groups
+
+==== Master
+A dedicated group is created to allow access to the following ports:
+
+22:: restricted to an optional parameter.
+Users would typically use `curl ident.me` to avoid making this port accessible to the world.
+8080:: not restricted.
+50000:: not restricted (for agents).
```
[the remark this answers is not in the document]
batmat: Yes. It would be an improvement. Though I'm finally not even sure we need that port, see below in the secu part. Need to check that.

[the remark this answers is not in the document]
batmat: OK, I just checked and this port is actually useless. I'm not that surprised as IIUC this goes through the same pipe as when using manually the [SSH Agent plugin](https://wiki.jenkins.io/display/JENKINS/SSH+Agent+Plugin), but I had never checked. Thanks for challenging it @dcoraboeuf. Going to fix this. Will make it more secure, and simpler in one go \o/.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+
+[source,json]
+"Statement": [{
+        "Sid": "TodoRefineSecurityALot",
+        "Effect": "Allow",
+        "Action": [
+            "s3:PutObject",
+            "s3:GetObject",
+            "s3:ListBucket",
+            "s3:DeleteObject"
```
jglick: You do not need this permission by default: unless overridden with a system property (which I suppose Essentials does not support?), the plugin will not delete artifacts or stashes from S3—that is something that (if even desired) should be configured in S3 itself.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+    "Resource": "*"
+}]
+
+==== S3 bucket policy
+
+FIXME: refine through JENKINS-52342.
+
+[source,json]
+"Statement": [{
+        "Sid": "TodoRefineSecurityALot",
```
jglick: ?

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+        "Resource": "*"
+    }
+]
+
+=== Security groups
+
+==== Master
+A dedicated group is created to allow access to the following ports:
+
+22:: restricted to an optional parameter.
```
jglick: So by default, no CLI over SSH, nor Git server for (old-style) `workflowLibs.git`, `userContent.git`, etc.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+Users would typically use `curl ident.me` to avoid making this port accessible to the world.
```
jglick: Can you clarify?

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+
+== Backwards Compatibility
+
+There are no backwards compatibility concerns related to this proposal.
+
+== Security
+
+=== Allowed certificates
+
+link:https://github.com/jenkinsci/jep/tree/master/jep/307#security[JEP 307] aggressively restricts the list of certificates that will be allowed to be used from inside Essentials container.
```
jglick: Seems like a poor design choice in JEP-307. It should not touch the JRE-wide certificate list. Rather, the one piece of code which actually retrieves an update manifest should specify the root certificates it trusts.

Anyway I thought that this transmission was to the Evergreen client rather than to Jenkins itself, so why would this have any effect on Jenkins plugins? This is not making sense to me.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+
+Two more need to be re-enabled for the Essentials AWS flavor to be able to auto-detect permissions given by the current instance profile:
+
+* `Baltimore_CyberTrust_Root.crt`: for S3.
+* `Amazon_Root_CA_1.crt`: for EC2.
+
+=== SSH access to the server
+
+Users are strongly advised to pass a specific IP when creating their setup.
+The `SSHLocation` template parameter should be a specific IP.
```
jglick: A specific IP of what? Please explain.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+=== AWS account for testing
+
+If we want to be able to run automated tests, we need some form of AWS account to actually check the CloudFormation works and keep working. See below <<aws-testing>>.
+
+== Testing
+
+[[aws-testing]]
+=== AWS testing of CloudFormation template
+
+Given an <<aws-account>>, we can use the `aws` CLI, we need to set up automated tests using the .
```
jglick: The what?

jep/0000/README.adoc
batmat: https://issues.jenkins-ci.org/browse/JENKINS-52342

jep/0000/README.adoc
batmat: Done.

jep/0000/README.adoc
batmat: We do not touch the JRE certificates, we modify the whole Docker image certificates. So it affects the whole system.

See https://github.com/jenkins-infra/evergreen/blob/a949d54575517f1e29885043bcff21617a70c6c9/distribution/scripts/generate-ca-certificates

jep/0000/README.adoc
batmat: Fixed thanks

jep/0000/README.adoc
batmat: > Seems like a poor design choice in JEP-307

Care to develop that statement? I think this is actually a good thing to aggressively reduce the set of targets we allow things to access from the master container. This would make it indeed impractical to run workload from the master and do anything from there (even https://github.com would be considered insecure), but that is actually almost a feature as we do **not** want to run things on the master.

jep/0000/README.adoc
rtyler: Everybody should care about deletions! There's cost, which I agree is cheap, but it's also a liability (IMHO) to unnecessarily maintain older artifacts, especially when developers may be inadvertently archiving data they shouldn't be (nobody ever archives secrets! :smiling_imp: )

I don't see a problem retaining the permission, regardless of the current state of the support by the s3 Artifact Manager plugin.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+* `Baltimore_CyberTrust_Root.crt`: for S3.
+* `Amazon_Root_CA_1.crt`: for EC2.
+
+=== SSH access to the server
+
+A CloudFormation template parameter `SSHLocation` is used to define the IPs allowed to access the master using SSH.
+
+Users are strongly advised to pass only their own IP when creating their setup.
+
+// Should we actually fail if one passes 0.0.0.0/0?
```
carlossg: I think SSH is safe enough, no problem opening it to the world

bitwiseman:
@batmat Is this sufficiently advanced to submit as draft?

batmat:
@bitwiseman sorry for being such a slowpoke. I've just re-read comments, and I think yes, we can merge as draft. All comments have been addressed AFAICT, be it by additional commit, or by answering to the given comment. Thanks!

jep/0000/README.adoc
batmat: So I'm finally leaving it as-is for a similar reason as the port `50000` allowed above: though the port is not yet used/enabled on the Jenkins Essentials master side itself, we still put that setting as such so that users can enable or disable it easily just by changing config in Jenkins itself.

Essentials is critically about simplicity, so here it's in my opinion the winning criterion to keep such settings this way, provided obviously we agree it's not a glaring wide security hole (as I think it is not, here).

Going to add a blurb about it in the spec to keep track of this, not just as a comment here.

jep/0000/README.adoc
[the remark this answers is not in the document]
batmat: @dcoraboeuf, oh, and thank you again. In case you or anyone wonder: your input was **very** valuable. Even deciding to close as WONTFIX is already an important improvement, instead of just doing things for no clear reason, your challenge pushed us to actually think more deeply about if and why we did want to keep that port open.

batmat:
Just added a blurb recaping https://github.com/jenkins-infra/evergreen/pull/131

jep/0000/README.adoc
batmat: > nor Git server for (old-style) workflowLibs.git

I _think_ the "old-style" part of the message sums it up. As per the [_Obvious Path To User Success_](https://github.com/jenkinsci/jep/tree/master/jep/300#obvious-path) pillar, I think we shouldn't enable this.

Or do you think there are still use cases that would be important for "modern" usages nowadays, and easy enough for the [target audience](https://github.com/jenkinsci/jep/tree/master/jep/300#target-audience) of Essentials?

PR #149, opened 2018-07-24
[JEP-207] - Copy-edit design decisions and body

oleg-nenashev:
CC @jglick @bitwiseman

### DIFF
--- a/jep/207/README.adoc
+++ b/jep/207/README.adoc
@@ -284,7 +284,7 @@ gets the entire log as a single String
 ** This method is deprecated in `hudson.model.Run`,
 and it should remain deprecated
 * `List<String> getLog(int maxLines) throws IOException` -
-gets a number of log lines as a
+gets a number of log lines as a list of strings
 * `File getLogFile() throws IOException` -
 Compatibility method, which retrieves the log as a `File`.
 ** By default a temporary file will be created, unless an implementation offers something better
--- a/jep/207/README.adoc
+++ b/jep/207/README.adoc
@@ -85,15 +85,17 @@ Support of other logging types will be documented in subsequent JEPs.
 * Minimize log traffic between Jenkins agents and the master, logs should be reported from slaves directly when it is possible
 * External Build Log Storage has an extensible and pluggable architecture
 * Reference implementations:
-   * File-system-based storage (current implementation)
-   * One external build log storage based on the open-source stack, based on Fluentd/Logstash and CloudWatch for AWS
-   * Nice2Have: Elasticsearch-Logstash-Kibana for self-hosted solutions
-* Jenkins Pipeline is supported out of the box
-* Real-time and Annotated run console logs should be still supported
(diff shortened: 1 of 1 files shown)


jep/207/README.adoc
```diff
@@ -85,15 +85,17 @@ Support of other logging types will be documented in subsequent JEPs.
...
+* Freestyle and Jenkins Pipeline projects are supported out of the box
+* Run console logs should be provided using standard Jenkins interfaces
+* Data migration flow for upgrading instances is technically possible
+** It is OK if it requires special actions outside Jenkins
 * Reference External Build logging implementation provides a feasible performance and fault-tolerance compared to the original filesystem based solution
-* All Global Configurations are designed in a way that they can be configured via Jenkins Configuration-as-Code Plugin OOTB[e][f]
-Presumptions, obstacles and limitations (braindump)
+* All Global Configurations are designed in a way that they can be configured via Jenkins Configuration-as-Code Plugin
+
+=== Design decisions
```
jglick: Better to follow the standard JEP structure: **Reasoning** and **Motivation**.

jep/207/README.adoc
```diff
@@ -249,17 +273,21 @@ Get large text for the entire execution/run
...
-get lof as a Reader
-* `String getLog() throws IOException`
-* `List<String> getLog(int maxLines) throws IOException`
+get the log as a Reader
+* `String getLog() throws IOException` -
+gets the entire log as a single String
+** This method is deprecated in `hudson.model.Run`,
+and it should remain deprecated
+* `List<String> getLog(int maxLines) throws IOException` -
+gets a number of log lines as a
```
jglick: as a…?

jep/207/README.adoc
oleg-nenashev: I really cannot read JEPs when decision making goes after the specification. I would prefer to keep it here unless it violates JEP requirements

bitwiseman: @oleg-nenashev 
The choice of top level sections and their order was discussed and agreed on.  It is set and should not be changed.     If it matters to you that it be another way, propose the change i
including bringing all other JEPs into compliance with the new structure.  


It was an mistake on my part as an editor that I didn't catch that this JEP does not meet the requirements.  I'll make the updates to bring it back into compliance.

bitwiseman: @oleg-nenashev  
Looking now, I see that the reasoning for the section ordering is not mentioned in the JEP-1, but I know it was discussed.  I will go dig up the discussion and make sure it is added to the Reasoning.

oleg-nenashev: Anyway, it is quite unrelated to this PR, it does not really change it.
So I would prefer to merge it as is and then discuss.

bitwiseman:
@oleg-nenashev 
I have merged this PR.  Please submit a new one to bring the JEP structure into compliance.

PR #147, opened 2018-07-23
New JEP: External Build Logging support in the Jenkins core

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-52691

@carlossg @jglick @jenkinsci/jep-editors

PR #151, opened 2018-07-24
New JEP: External Logging API Plugin

oleg-nenashev:
Follow-up to JP-207, which defines the API layer to be reused by other implementations.

@jenkinsci/jep-editors @jglick  @carlossg

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -48,7 +48,7 @@ endif::[]
 //
 // Uncomment if this JEP depends on one or more other JEPs.
 | Requires
-| link:/jep/jep-207[JEP-207]
+| link:/jep/jep-207[JEP-207], link:/jep/jep-210[JEP-210]
 //
 //
 // Uncomment and fill if this JEP is rendered obsolete by a later JEP
@@ -211,22 +211,45 @@ As a reference implementation of the _External Logging API_ plugin,
 a new _External Logging for Elasticsearch Plugin_ will be implemented.
 Other implementations may be also created.
 
-== Reasoning
+== Motivation
 
-=== Why separate plugin?
-
+JEP-207 introduces a new API in the core for adding External logging features,
+but it does not provide neither configuration UI nor convenient API for implementing these storage engines.
+This API plugin does that.
 One may say that all bits in the current design could be implemented
 as a part of the Jenkins core.
-It is true, but detaching of the plugin is intentional:
+It is true, but detaching of the plugin has the following motivation:
 
 * The plugin will have a separate release cycle so that changes in it
   can be delivered and backported independently from the core's release cycle
 * The approach allows keeping the patches on the core's side minimal
+* The approach allows integrating with Pipeline Log Storage API introduced in Pipeline plugins (JEP-200)
 
 All External Log Storage implementations are expected to extend this plugin
 instead of just using API provided by the Core.
 Core APIs may be still used to define custom `LoggingMethodLocator` impelemtations,
 e.g. to define a custom logger allocation logic.
+
+== Reasoning
+
+=== Why do we introduce the Event layer?
+
+Jenkins project usually operates with logs as data streams and lines,
+especially on the agent side.
+On the other hand, modern log storage systems operate with "events" -
+atomic objects which may include multi-line strings and various metadata.
+Example: exception stacktraces may go to log storage as a single event and
+then they can be processed by external systems like Logstash if needed.
+The idea in this plugin is to offer bridge logic which converts
+stream-based logging into event-based logic.
+
+Jesse Glick has raised the concern that Event layer may not be helpful
+taking the current state,
+because a lot of code would need to be updated so that the events get captured properly.
+Opinion of the JEP sponsor is that this JEP offers a foundation layer
+so that it may be implemented.
+Reworking the entire Jenkins API to events is NOT an objective for this JEP-200,
+but it may be added in subsequent JEPs.
 
 === Why do we need JSON layer?
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -32,7 +32,7 @@ endif::[]
 | 2018-07-24
 
 | BDFL-Delegate
-| :bulb: Link to github user page :bulb:
+| TBD
 
 //
 //
--- a/jep/0001/README.adoc
+++ b/jep/0001/README.adoc
@@ -168,8 +168,11 @@ It will be sent to the agent side.
 === ExternalLogBrowserFactory and classes
 
 This factory just produces instances of `ExternalLogBrowser`.
-This class offers an abstraction layer for external log browsing
-and so far does not implement any custom methods.
+This class offers an abstraction layer for external log browsing.
+
+`ExternalLogBrowser` will also provide an abstraction layer for
+eventual consistency management.
+This layer will be determined during reference implementation polishing.
 
 === JSON-based external-logging layer
 
@@ -182,6 +185,8 @@ This layer should provide the following features:
 * Utility classes for reading and writing JSON Events,
 including parsing/writing `ConsoleNote` objects
 * Base classes for constructing JSON queries for fetching data
+
+Other convenience layers will be defined during prototyping.
 
 === Configuration
 
@@ -222,6 +227,21 @@ Many popular log storage engines store events in a JSON format:
 Fluentd, Logstash, Elasticsearch, AWS CloudWatch, etc.
 Offering a JSON layer as a part of the API plugin could greatly
 simplify such implementations.
+
+=== Eventually-consistent storages
+
+Some target storages are eventually consistent.
+One cannot just write the data to remote storage and then reliably read it.
+It is critical for log browsers:
+
+* When a run finishes, querying data does not guarantee we get all the data
+* `Loggable#isCompleted()` call is not enough,
+some entries may be missing for "completed" entries
+
+Such issue explains why we need a special watch layer to determine whether logs are actually completed.
+It explains why we need a special logic/queries to determine whether the log is actually complete.
+
+Full logic for the layer should be determined during reference implementation.
 
 == Backwards Compatibility
 
--- a/jep/0001/README.adoc
+++ b/jep/0001/README.adoc
@@ -1,0 +1,279 @@
+= JEP-0000: External Logging API plugin
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| JEP
+| 0000
+
+| Title
+| External Logging API Plugin
+
+| Sponsor
+| link:https://github.com/oleg-nenashev[Oleg Nenashev]
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| Standards
+
+| Created
+| 2018-07-24
+
+| BDFL-Delegate
+| :bulb: Link to github user page :bulb:
+
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+//| JIRA
+//| :bulb: https://issues.jenkins-ci.org/browse/JENKINS-nnnnn[JENKINS-nnnnn] :bulb:
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
+//
+//
+// Uncomment if this JEP depends on one or more other JEPs.
+| Requires
+| link:/jep/jep-207[JEP-207]
+//
+//
+// Uncomment and fill if this JEP is rendered obsolete by a later JEP
+//| Superseded-By
+//| :bulb: JEP-NUMBER :bulb:
+//
+//
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+On large-scale Jenkins instances master Disk and Network I/O become bottlenecks in particular cases.
+Build logging and Task reporting are one for the most intensive I/O consumers,
+hence it would be great to somehow redirect them to an external system.
+This is a continuation of the original story we had back in 2016
+(see the public design document
+link:TODO[here]).
+JEP-TODO defines requirements to the external logging system and Core APIs.
+
+The External Logging API plugin is a separate plugin
+which offers shared functionality to External Logging implementations.
+
+* Configuration UI
+* Extension Points for External Logging implementations
+* Base classes for external logging,
+which simplify implementations
+* Other shared utility classes
+
+== Specification
+
+=== Event class
+
+Event is an atomic entry of the external logging.
+In the data model it includes the following fields:
+
+* `id` (long) - ID of the message.
+  It must be unique within a single `LoggingMethod` writer
+* `message`(string) - Contains the message body.
+  It may be a single-line or a multi-line message.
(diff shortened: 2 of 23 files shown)


jglick:
Why is this a separate JEP? It seems to me like it should be folded into 207, so that we are describing a complete and self-contained user story.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+* `id` (long) - ID of the message.
+  It must be unique within a single `LoggingMethod` writer
+* `message`(string) - Contains the message body.
+  It may be a single-line or a multi-line message.
+  The message may also contain annotations.
+* `timestamp` (long) - Timestamp of the message,
+  measured in milliseconds, between
+  the current time and midnight, January 1, 1970 UTC
+* `data` (`Map<String, Serializable>`) - additional metadata.
+  This metadata will be used to categorize event entries.
```
jglick: This does not make sense to me. How could you portably describe what these objects are? AFAICT in the reference implementation these are always `String`s (but are declared to be `Object`s).

oleg-nenashev:
> Why is this a separate JEP? It seems to me like it should be folded into 207, so that we are describing a complete and self-contained user story.

Just because JEP-207 was getting too large + this JEP may be affected by the Pipeline log storage design. It may also take longer to get this JEP reviewed, because it is more complicated in terms of the implementation. I would prefer to keep them separate for now.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+* `loggableType`  - Type of the `Loggable` entry, e.g. `run`
+* `loggableId` - Identifier of the `Loggable` entry
+
+==== Run metadata
+
+The following metadata entries are required to be present
+for all `hudson.model.Run` event entries:
+
+* All event metadata
+* `jobId` - ID of the Job, should be provided by Unique ID Plugin
```
jglick: Not really desirable as it is opaque to external tooling. Would recommend using `Item.fullName`, and optionally implementing an `ItemListener.onLocationChanged` allowing an external provider to either update external storage or retain a redirect field.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+Requited fields are defined below.
+
+==== Event metadata
+
+The following metadata entries are required to be present
+for all event entries:
+
+* `jenkinsId` - Instance ID of the Jenkins instance
+* `loggableType`  - Type of the `Loggable` entry, e.g. `run`
+* `loggableId` - Identifier of the `Loggable` entry
```
jglick: Why would this not be the same as `jobId` plus a build number?

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+
+Some metadata is mandatory.
+Requited fields are defined below.
+
+==== Event metadata
+
+The following metadata entries are required to be present
+for all event entries:
+
+* `jenkinsId` - Instance ID of the Jenkins instance
```
jglick: This is going to add a bunch of overhead that most installations do not need.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+* `loggableType`  - Type of the `Loggable` entry, e.g. `run`
+* `loggableId` - Identifier of the `Loggable` entry
+
+==== Run metadata
+
+The following metadata entries are required to be present
+for all `hudson.model.Run` event entries:
```
jglick: No build number?

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+
+* All event metadata
+* `jobId` - ID of the Job, should be provided by Unique ID Plugin
+
+==== Pipeline metadata
+
+* All Run metadata
+* `stepId` - Identifier of the step ("node")
```
jglick: Recommend `node`, since this is not exactly a “step”. (Could be for example a particular body block.)

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+`ExternalLogBrowser` will also provide an abstraction layer for
+eventual consistency management.
+This layer will be determined during reference implementation polishing.
+
+=== Pipeline Bridge extensions
+
+The plugin should also implement Pipeline `LogStorage` and `LogStorageFactory` extension points
+so that it transparently supports Pipeline with existing API.
+
+Pipeline Storage JEP is documented in JEP-TODO.
```
jglick: coming shortly

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+=== JSON-based external-logging layer
+
+The API Plugin should offer a convenience layer in order
+to support a number of most common logging providers like
+Logstash, Fluentd, Elasticsearch, AWS CloudWatch, etc.
+
+This layer should provide the following features:
+
+* Utility classes for reading and writing JSON Events,
+including parsing/writing `ConsoleNote` objects
```
jglick: This is not really specific to JSON. Would apply to any log sink accepting structured records of any format.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+
+=== Why separate plugin?
+
+One may say that all bits in the current design could be implemented
+as a part of the Jenkins core.
+It is true, but detaching of the plugin is intentional:
+
+* The plugin will have a separate release cycle so that changes in it
+  can be delivered and backported independently from the core's release cycle
+* The approach allows keeping the patches on the core's side minimal
```
jglick: Is it really minimal? Minimal core APIs would be some extension point just allowing a plugin to override the behavior of certain critical methods.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+It is critical for log browsers:
+
+* When a run finishes, querying data does not guarantee we get all the data
+* `Loggable#isCompleted()` call is not enough,
+some entries may be missing for "completed" entries
+
+Such issue explains why we need a special watch layer to determine whether logs are actually completed.
+It explains why we need a special logic/queries to determine whether the log is actually complete.
+
+Full logic for the layer should be determined during reference implementation.
```
jglick: BTW this is solved for the Pipeline implementation.

jep/0001/README.adoc
```diff
@@ -0,0 +1,306 @@
...
+
+=== Functional testing
+
+All tests will be implemented using Jenkins Test Harness or
+Acceptance Test Harness (ATH) frameworks.
+
+The following use-cases must be covered:
+
+* Backward compatibility
+* Upgradeability - upgraded instances use the Filesystem Storage by default
```
jglick: Upgradability

jglick:
> It may also take longer to get this JEP reviewed, because it is more complicated

Exactly why it makes no sense to review 207 in isolation—it is worthless without all these other changes.

jep/0001/README.adoc
oleg-nenashev: sure, should be added

jep/0001/README.adoc
oleg-nenashev: `node` is going to confuse everybody (agent vs. Pipeline graph). I decided to use "stepId", because it allows doing some grouping for non-Pipeline logs if needed

jep/0001/README.adoc
oleg-nenashev: It's pretty much what JEP-207 does. `LoggingMethod` + `LogBrowser` + `LoggingMethodLocator`

jep/0001/README.adoc
oleg-nenashev: For sure.

jep/0001/README.adoc
oleg-nenashev: For example, if we have some TaskLogging in the future. They will be going through the same engine, so I want to keep required changes minimal

jglick:
cf. #147 & #153

oleg-nenashev:
@jglick @bitwiseman As a JEP sponsor I believe it should be a separate JEP. JEP-207 may be approved separately in such case, which is important IMO

Apart drom the technical comments, I see no other unadressed comments. Can we merge it as a draft?

bitwiseman:
@oleg-nenashev Missing required motivation section.

oleg-nenashev:
@bitwiseman done

oleg-nenashev:
@jglick. Could you please communicate design concerns in rhe ML? I will be
happy to follow-up in next PRs, but rhis one is about form, not the the
final design.

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+There is no special log testing requirements for this story.
+External Logging API and its implementations are responsible to execute
+performance and load testing,
+if deemed necessary.
+
+== Prototype implementation
+
+* https://github.com/oleg-nenashev/external-logging-api-plugin
+* https://github.com/oleg-nenashev/external-logging-logstash-plugin
```
jglick: These got forked to @jenkinsci so you should update the URLs.

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+The following metadata entries are required to be present
+for all `hudson.model.Run` event entries:
+
+* All event metadata
+* `jobId` - ID of the Job, should be provided by Unique ID Plugin
+
+==== Pipeline metadata
+
+* All Run metadata
+* `stepId` - Identifier of the step ("node")
```
svanoort: I think you probably mean flowNodeId, since that's the actual object that is mapped to where a log originates from (1 or more flownodes per step - an important key point).

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+* `id` (long) - ID of the message.
+  It must be unique within a single `LoggingMethod` writer
+* `message`(string) - Contains the message body.
+  It may be a single-line or a multi-line message.
+  The message may also contain annotations.
+* `timestamp` (long) - Timestamp of the message,
+  measured in milliseconds, between
+  the current time and midnight, January 1, 1970 UTC
+* `data` (`Map<String, Serializable>`) - additional metadata.
```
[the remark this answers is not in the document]
svanoort: Better yet:  JSON (or Map<String, JsonType> that gets converted to a JSON document)  for most things, and a Map<String, Serializable> of JenkinsInternal stuff -- with as little as possible in the latter category. 

To do cloud-native effectively, we need to have something consumable directly from the browser or that potentially be consumed by a non-Java source.  (ab)use of java serialization is generally discouraged at this point -- we should only be doing it where 100% required.

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+* `loggableType`  - Type of the `Loggable` entry, e.g. `run`
+* `loggableId` - Identifier of the `Loggable` entry
+
+==== Run metadata
+
+The following metadata entries are required to be present
+for all `hudson.model.Run` event entries:
+
+* All event metadata
+* `jobId` - ID of the Job, should be provided by Unique ID Plugin
```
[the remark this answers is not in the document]
svanoort: We need *some* sort of identifier but I echo Jesse's concerns.  This is kind of a damned-if-you-do, damned-if-you-don't scenario around renames.  

Where I differ is that I think we **do** need a unique-id, but that needs to get extended to Multibranch as well (however I'm very okay with Multibranch using a unique id for the multibranch project, with the branch name concatenated to form a complete ID).  Would propose doing that - I'd imagine not a big amount of work in all likelihood.

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+Event is an atomic entry of the external logging.
+In the data model it includes the following fields:
+
+* `id` (long) - ID of the message.
+  It must be unique within a single `LoggingMethod` writer
+* `message`(string) - Contains the message body.
+  It may be a single-line or a multi-line message.
+  The message may also contain annotations.
+* `timestamp` (long) - Timestamp of the message,
```
svanoort: Timestamp when first part of message generated, or last?  Or perhaps both?

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+ExternalLoggingEventWriter class:
+
+* The class is `Serializable`.
+It will be sent to the agent side.
+* The class offers the following abstract methods:
+** `void writeEvent(Event event) throws IOException`
+- writing event to the remote storage
+* The class also stores metadata, which may be injected into the events
+** The class stores a Map of Serializable metadata entries
```
svanoort: Similarly, consider a non-Java map of metadata that can be consumed by other sources, plus a smaller Java serialized grouping.

jep/0000/README.adoc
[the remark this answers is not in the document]
svanoort: Also, wouldn't jobId just by the `loggableId` for the 'Job' `loggableType` above?

jep/0000/README.adoc
svanoort: This makes parsing easier for non-java consumers too

jep/0000/README.adoc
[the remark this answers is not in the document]
svanoort: Actually yeah, here's my proposal: we have JSON field that can be trivially converted by Java since types will map over.   Then we have a Map<String, Serializable> with serializables stored in hex, which can be ignored by non-Java consumers.

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+The API Plugin should offer a convenience layer in order
+to support a number of most common logging providers like
+Logstash, Fluentd, Elasticsearch, AWS CloudWatch, etc.
+
+This layer should provide the following features:
+
+* Utility classes for reading and writing JSON Events,
+including parsing/writing `ConsoleNote` objects
+* Base classes for constructing JSON queries for fetching data
```
svanoort: Consider GraphQL for an event querying API?

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+Some target storages are eventually consistent.
+One cannot just write the data to remote storage and then reliably read it.
+It is critical for log browsers:
+
+* When a run finishes, querying data does not guarantee we get all the data
+* `Loggable#isCompleted()` call is not enough,
+some entries may be missing for "completed" entries
+
+Such issue explains why we need a special watch layer to determine whether logs are actually completed.
```
svanoort: Polling-based or notification-based watch layer?  This strikes me as a wrinkle that may be deceptively complex.

oleg-nenashev:
@svanoort could you please submit your comments in the mailing list? JEP draft PRs are not designed for design discussions (though pretty much everybody gets confused by that).

bitwiseman:
@svanoort 
I'm sorry for merging over your objections.  I simply didn't seem them - forgot to refresh. 

That said I think we (as a community) need to talk about a bit more about how to use PRs for approval as Draft.  Currently, it seems the a large part of the design discussion is occurring during the initial PR.  This PR shows one reason why this is problematic.  The one time a JEP document moves is during approval as draft, so comments (while not lost) do end up being hidden.  From the flip side, the intent of draft approval is to say "This JEP is sufficiently _started_.  Now design discussion can continue."

I have opinions and thoughts around this, but I'm also very open and interested in figuring out a working model for this that serves everyone's needs while not being an overburdening process. 

It would be great to chat about this more at this week's JEP Office hours.  I

svanoort:
@oleg-nenashev @bitwiseman No worries, and I do think there's a discussion to have about where to do initial JEP feedback. 

> it seems the a large part of the design discussion is occurring during the initial PR

I've noticed this too, and IMO that's actually a good thing -- it lets the sponsor and community iteratively refine the proposal and consolidates a lot of the initial discussion.  

IMO the draft state was a good *idea* to ensure there's a consistent proposal, but we hit challenges with the feedback process, because the jenkins-dev list isn't as elegant (and follow-up PRs as a way to gather feedback or propose changes are really clunky and fragment the discussion).  From a practical standpoint the interface and notifications for Google Groups are not as convenient or refined as the Github issues/PRs system, **especially** where we need to reference specific lines of content under discussion.  It's also more easy to find discussion as a matter of historical record, since they're tied to the content.

I'd propose that we convert draft state to something more like a semifinal proposal, with yes-or-no discussion happening on the dev list (and room to still submit revisions based on that discussion).

Yes, that means things stay in draft longer, but they'll have a much faster process once they leave draft, and we're using JEPs for substantial changes anyway.

PR #153, opened 2018-07-24
[JENKINS-38381] Draft JEP

jglick:
Submitted for approval as draft JEP by @jenkinsci/jep-editors. CC @oleg-nenashev @carlossg @svanoort @abayer @rsandell @dwnusbaum

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -347,7 +347,8 @@ For now, these considerations were outweighed by the convenience of running on s
 When the synchronous `Launcher` interface is used to start non-durable remote processes,
 as happens for example from typical `SCM` implementations delegating to a command-line tool,
 currently the remotability of any supplied `TaskListener` is ignored
-and all log lines are sent over the Remoting channel to be processed on the master side.
+and all log lines are sent over the Remoting channel to be processed on the master side:
+link:https://issues.jenkins-ci.org/browse/JENKINS-52729[JENKINS-52729].
 This is likely fixable as a simple patch to `Launcher`,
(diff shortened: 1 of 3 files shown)


svanoort:
I do not understand why this would be a JEP?  The parts that seem JEP-appropriate are well-covered by JEP-207 and https://github.com/jenkinsci/jep/pull/151 and the parts that are not there are simply implementation details.

Implementations are supposed to be handled via normal code review process.  Or are we assuming that every single code change we do should have a JEP associated with it?

svanoort:
Actually... I'll go a step further than that: explaining implementation details *should* go into the repo with the code change -- either as docs files or code comments.  And I'd really like to see those code changes get finished.

jglick:
JEP-207 + #151 covers general changes including for freestyle; I expect these to be reconciled.

The specific topics in Pipeline covered here go far beyond implementation details or simple bug fixes and merit the structure of a JEP. This was confirmed recently in the JOM by @bitwiseman and others.

oleg-nenashev:
+1. I believe it should be a JEP. It is a major API enhancenent for
Pipeline users

oleg-nenashev:
#151 actually depends on this draft  in the current design (Pipeline logging bridge methods)

jglick:
And of course individual PRs do add Javadoc for newly introduced API members, which is the authoritative source of information about fine details like that. I make no attempt to duplicate all of that content here, only highlight the more subtle or critical aspects. The JEP is here to describe the higher-level _purpose_ of the entire assembly, within the context of one or more reference implementations, and should be the permanent record of that overall design.

PR #154, opened 2018-07-27
JEP table: there may be multiple sponsors

oleg-nenashev:


bitwiseman:
@oleg-nenashev To update this you need to update the script that generates it - `enumerate-jeps`.

PR #159, opened 2018-07-31
Added comments/questions from svanoort

bitwiseman:
@oleg-nenashev 
These were the comments from @svanoort  on #151 .  I believe you can respond by editing or commenting directly on this PR (if not please tell me so we can figure out how to make that possible).   

@jglick  I believe you also had some comments/edits.  Feel free to make them here or separately.

As noted in #151, many people have take the initial submission as invitation to do a full review.  That it explicitly not how it was described in the JEP-1.  This PR is partly an experiment to explore how we can make draft submission PRs less like design reviews while still addressing the point that people want to be able to comment on JEPs. This is one way I thought we could work around the GitHub limitations on discussion/questions/suggestions outside of PRs.

### DIFF
--- a/jep/212/README.adoc
+++ b/jep/212/README.adoc
@@ -96,8 +96,22 @@ In the data model it includes the following fields:
 * `timestamp` (long) - Timestamp of the message,
   measured in milliseconds, between
   the current time and midnight, January 1, 1970 UTC
+////
+(svanoort)
+Timestamp when first part of message generated, or last? Or perhaps both?
+////
 * `data` (`Map<String, Serializable>`) - additional metadata.
   This metadata will be used to categorize event entries.
+////
+(svanoort)
+Why not enforce something simpler than Serializable for (meta)data which can be consumed by non-Java clients potentially, i.e. JSON perhaps?
+
+Better yet: JSON (or Map<String, JsonType> that gets converted to a JSON document) for most things, and a Map<String, Serializable> of JenkinsInternal stuff -- with as little as possible in the latter category.
+
+To do cloud-native effectively, we need to have something consumable directly from the browser or that potentially be consumed by a non-Java source. (ab)use of java serialization is generally discouraged at this point -- we should only be doing it where 100% required.
+
+Actually yeah, here's my proposal: we have JSON field that can be trivially converted by Java since types will map over. Then we have a Map<String, Serializable> with serializables stored (edited)base64, which can be ignored by non-Java consumers.
+////
 
 Some metadata is mandatory.
 Requited fields are defined below.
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
> These were the comments from @svanoort on #151 . I believe you can respond by editing or commenting directly on this PR (if not please tell me so we can figure out how to make that possible).

Looks like an attempt to replicate Google Doc.

> This is one way I thought we could work around the GitHub limitations on discussion/questions/suggestions outside of PRs.

I also have no idea how to do it better tho.
If Google Doc (or something like that) supported import/export to Asciidoc, it would be easier to do something with that, even to setup automated flow somehow

oleg-nenashev:
I tried several external converters. Didn't work well

bitwiseman:
Also, see #160 for an example using visible blocks instead of comments.

bitwiseman:
@oleg-nenashev
One difference being the history is better tracked (and not deletable) in GitHub. 

And yes, I wish there were a Google doc export to asciidoc. :|

bitwiseman:
https://asciidoctor.org/docs/migrating-from-msword/ suggests a bunch of ideas, but none are great.

svanoort:
This is a clever approach for draft JEPs.  I think I prefer this to #160 because it doesn't clutter up the display out-of-box.  Um, I can't edit directly though -- probably because I'm not a sponsor or BDFL-delegate at the moment.

Apparently you can embed HTML into ASCIIdoc so there's room to embed scripts for expandable content blocks, i.e. "see discussion" etc.   But on the other hand, I think that is to some extent reinventing the wheel by trying to re-make Google Docs (agree with @oleg-nenashev  there). 

@bitwiseman There's a value to this technique for proposed JEPs, and/or where we want to permanently note some incidental feedback raised (also #160 where it's important it be visible when viewing the JEP).   But I still think Github PR review tooling is the best way to do the early stages.  It's just *so* much faster to iterate using that while incorporating feedback. 

@oleg-nenashev @bitwiseman My proposal would be to convert "draft" status -> "proposed" JEP status.  A proposed JEP would meet all the current criteria for a draft, but additionally should have more or less stabilized.  The sponsor would ask to go ahead into 'proposed' state if they've incorporated as much feedback as they feel able to.  JEP editors might poke them to do so when it looks like the discussion & proposal changes have settled down a bit -- that doesn't mean everyone _agrees_, of course, because there will always be compromises.

A proposed JEP would go up for discussion on the dev-list (as currently) and could still be amended based on that, but might also get accepted by the BDFL-delegate as-is if feedback is positive. 

In normal code reviews, think of a proposed JEP as equivalent to code put up for final review, the PR state preceding it as a WIP PR, and dev-list discussion as final code review votes (with the BDFL-delegate taking the role of maintainer).

bitwiseman:
@svanoort 
"permanently note some incidental feedback raised" - That is what the Reasoning section is for. 

As to your proposal - What you are suggesting is a rather significant change.  But if you are passionate about this, you should feel free to raise this on the dev list, gather feedback, make a PR to JEP-1, drive this to completion.   Or, using the model shown in this PR (or #160), you can create a PR with your idea above in a block and people can discuss it there.

bitwiseman:
@svanoort 
Even if you cannot make direct edits you can create PR's against my PR branch - we used that model during JEP-1.  But for simplicity, I've made you a collaborator on my fork, so you should now be able to make direct edits.

bitwiseman:
Closing in favor of #160. We'll see how that goes.

PR #160, opened 2018-07-31
Added comments/questions from svanoort

bitwiseman:
@svanoort @oleg-nenashev @jglick 
This is an alternative to #159 . 

Either of these formats could be addressed by responding with comment or edits or the blocks could be  merged for later editing/response.  They would make it possible to have some degree of design discussion within the JEP (contained in the history) but also retain the final form of the JEP a design document (not a discussion thread).

This one as the advantage that questions/comments visible in the document.  It would make it easy to see  what the open questions/discussion are and be able to respond to them. The other format would make the JEP look more like it's final form (hidding comments/questions) from top level view. 

We could of course also use both as people felt appropriate.

### DIFF
--- a/jep/212/README.adoc
+++ b/jep/212/README.adoc
@@ -99,6 +99,24 @@ In the data model it includes the following fields:
 * `data` (`Map<String, Serializable>`) - additional metadata.
   This metadata will be used to categorize event entries.
 
+[WARNING]
+====
+*(svanoort)*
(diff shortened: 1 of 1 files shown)


bitwiseman:
@svanoort  
I think the easiest way to track if there our open concerns is by making them visible. 😄 No way to overlook them then.   

We could also update the enumerate-jeps to have a column indicating open questions/comments.

PR #162, opened 2018-08-02
Allow referencing particular threads for the JEP discussion.

oleg-nenashev:
It may be useful, especially for the Dev List with hundreds of threads. It is a follow-up to the discussion with @bitwiseman in JEP-4 (or other JEP) where I tried to reference the thread.

@jenkinsci/jep-editors

### DIFF
--- a/jep-template/0000/README.adoc
+++ b/jep-template/0000/README.adoc
@@ -62,7 +62,8 @@ If no suitable BDFL-Delegate can be found, that row may be commented out.
 //| :bulb: https://issues.jenkins-ci.org/browse/JENKINS-nnnnn[JENKINS-nnnnn] :bulb:
 //
 //
-// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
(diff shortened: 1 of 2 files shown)


bitwiseman:
I'm not a huge fan of this.  Linking to specific discussion threads is for References.  And pointing to one thread for discussion will probably result in a very long and wide ranging single thread rather than threads about specific parts of the JEP.  

But that is only my opinion, this should really be raised to a wider audience.

oleg-nenashev:
@bitwiseman Agreed. I have created a Label for such feedback, WDYT?

bitwiseman:
Label sounds like a good idea.  Thanks!

bitwiseman:
@oleg-nenashev 
Unless you plan to drive this forward, we should resolve this suggestion - either by adding a section to the Reasoning saying this was considered and passed over, or using a JIRA or other method to mark it for later consideration.  Open PRs are not the place to track future work.

bitwiseman:
@oleg-nenashev  
There's been no forward progress on this. Closing.  We can reopen if it comes up again

PR #166, opened 2018-08-06
Updated JEP-210

jglick:
Moved a “reasoning” block from https://github.com/jenkinsci/workflow-job-plugin/pull/27. Also adjusting wording to reflect the fact that https://github.com/jenkinsci/jenkins/pull/3563 was merged.


oleg-nenashev:
For me it looks like that "Reasoning" section has no explicit design decision

PR #172, opened 2018-08-08
JEP Draft: Configuration Storage API in the Jenkins Core

oleg-nenashev:
As discussed before, I have polished my writeup about External Configuration storage and reformatted it as JEP.

CC @deepy @jglick @carlossg


oleg-nenashev:
Also CC @jstrachan since it involves the Kubeify codebase

bitwiseman:
This looks like a Draft to me.    JEP-213.

PR #173, opened 2018-08-08
Use of non-ASCII-embedding encodings

jglick:
@svanoort asked about this topic in https://github.com/jenkinsci/durable-task-plugin/pull/60#discussion_r208308043.


oleg-nenashev:
LGTM

oleg-nenashev:
Merging so that it does not hang

PR #176, opened 2018-08-14
Updates based on community feedback.

tracymiranda:
Improvements based on community feedback


bitwiseman:
@oleg-nenashev Why not just comment here?  Or make an edit-comment in the way we discussed with svanoort?

PR #178, opened 2018-08-14
Updated JEP-210 for StreamLogStorage → FileLogStorage

jglick:
Matches recent changes in https://github.com/jenkinsci/workflow-api-plugin/pull/17.

### DIFF
--- a/jep/210/README.adoc
+++ b/jep/210/README.adoc
@@ -121,25 +121,40 @@ allowing an external log sink to bypass all master communications.
 
 === Replace aggregation with single log stream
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Should I merge it? Or do we wait for @svanoort 's feedback?

jglick:
@oleg-nenashev you can go ahead merge it—the PRs in the reference implementation have already been amended,and anyway something like this change was at @svanoort’s request to begin with. I had added him as a reviewer as a courtesy / heads-up.

PR #179, opened 2018-08-15
Rename from Jenkins Essentials to Jenkins Evergreen

rtyler:
See [this thread](https://groups.google.com/d/msgid/jenkinsci-dev/20180814220448.GI17800%40grape.lasagna.io)


rtyler:
aye

bitwiseman:
Could we change to evergreen.yaml, but support essentials.yaml as a deprecated name?  Considering Essentials hasn't hit v1.0, now would be the time to make the change.

rtyler:
@bitwiseman there are already three tools which I know of relying on `essentials.yaml` which I do not want to waste time changing here

PR #185, opened 2018-08-25
Add JEP for improving process for adopting a plugin

rodrigc:
Submitted for approval as draft JEP by @jenkinsci/jep-editors

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,102 @@
+= JEP-0000: Process improvements for adopting a Jenkins plugin
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.**JEP Template**
+
+.Metadata
+[cols="1h,1"]
+|===
+| JEP
+| 0000
+
+| Title
+| Process improvements for adopting a Jenkins plugin
+
+| Sponsor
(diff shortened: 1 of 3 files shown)


jep/0000/README.adoc
```diff
@@ -0,0 +1,102 @@
...
+* The new maintainer will be responsible for updating the wiki page for the plugin to list themselves as the new maintainer.
+* The new maintainer must updat the pom.xml metadata for the plugin in GitHub to indicate that they are the new maintainer.
+
+== Motivation
+
+Due to the popularity of Jenkins, the mailing lists receive a lot of messages.
+For casual Jenkins users who maintain plugins, following the mailing lists
+may be difficult and it is easy to miss these types of e-mails.
+
+== Reasoning
```
jglick: A lot of minor refinements or alternative structures were discussed on the dev list. You might not agree with any of them, but the basic ideas ought to be noted in this section and an explanation given of why you are going with the current proposal instead.

jep/0000/README.adoc
```diff
@@ -0,0 +1,102 @@
...
+a plugin.  This process must have a clear audit trail that can be tracked.
+There also must be a clear way to track the post plugin adoption tasks.
+
+== Security
+
+There are no security risks related to this proposal.
+
+== Infrastructure Requirements
+
+* JIRA must be updated to support the new *adopt-plugin* tag.
```
jglick: There is no preparation needed. The first time a new tag is typed in, it becomes available in completion.

bitwiseman:
This JEP looks rather skinny and needs quite a bit more information, but it meets the bar for Draft status.

I've include feedback from jglick in the JEP as WARNING sections.  
@daniel-beck  If you have feedback you can submit a pr with WARNING sections as well.

PR #191, opened 2018-08-29
Add Localization jep submission

LinuxSuRen:


oleg-nenashev:
Why don't we just repurpose the Translation plugin tho? It's already there, and it looks like a good option for an API plugin if we do some cleanup.

LinuxSuRen:
I think Translation plugin is just a tool for creating localiztion files, it doesn’t contain any resources file.

daniel-beck:
Two notable problems need to be addressed here:

* How does the translation keep up to date with substantial changes in the source material? Today, at least in core, if we substantially change a description or help text, we delete the localizations, as we consider a correct English description to be preferable to a wrong translated one. (It also signals the need to re-translate it.)
* Multiple versions of core or a given plugin are in use across the user base at any time. They might have different features, and they might be exposed in a way that means corresponding changes to the translations are needed -- if I add an option, I might mention the option in the help text for a related, older, feature. Wouldn't updated translations potentially mention or describe features that don't exist (yet) in the actual plugin? There should be a facility to expose what version of core or plugin a given description refers to, as well as a facility to switch from the translation to the original.

LinuxSuRen:
I offer a solution. Make a fallback mechanism, find the current version of the plugin or core of localization resources, if don't exists then find the latest version. Most of the cases just need to maintain the latest version.

daniel-beck:
@LinuxSuRen To clarify, a user might be running a plugin in version 4.3.2. Another user might be running version 4.5.6. You always attempt to localize for the latest plugin release, in this example 4.5.6.

Now if a feature was added between 4.3.2 and 4.5.6, there will be documentation for it. Not even just standalone in its own help file etc. (which would simply not be displayed), but there could be other documentation that says:

> This frobnicates a widget.

and is updated to:

> This frobnicates a widget unless the New Feature flag is enabled.

The user on plugin 4.3.2 would get a help text describing a feature that does not exist!

Some help text includes a note which plugin it is from (search for `X-Plugin-From` in the core code base). I wonder whether something similar can be done here, with a note

> "This help text was written for X plugin 4.5.6"

LinuxSuRen:
Sounds great. If it can't find matches version, then add a note like below.

bitwiseman:
@LinuxSuRen  I don't want to say that there is a minimum length on JEPs but this JEP is extremely short. I don't feel I can make sense of what you're proposing from the content of this JEP.  Please incorporate feedback from the ML and from this PR.

bitwiseman:
@LinuxSuRen @daniel-beck @slide @jglick 
I'm making this JEP-216 Draft so it can be referenced and discussed, but I want to reiterate, this JEP needs work.  It is barely a placeholder.  
@LinuxSuRen 
I would like to request that you make it a priority of the Chinese Localization SIG to get more detail in here.  If you please.

LinuxSuRen:
Thanks to everybody, this is my first JEP. I have a lot of things need to catch up. Sure there are many details is needed.

daniel-beck:
Sorry for the lack of responses recently. I was busy with JW, then with security updates. I really like where this is going. Central localization plugins should make it much easier to have consistent translations.

PR #197, opened 2018-09-11
[JEP-214] Nominate rtyler for BDFL Delegate

daniel-beck:


kohsuke:
The change in the core is pretty modest, and the majority of the work will be on the infrastructure side, and he wrote that app, so I think it makes sense that he is the BDFL delegate. I also think this and the Evergreen telemetry system should leverage the commonality as much as possible --- they both establish a feedback cycle for developers.

I note that this PR was merged too early. I appreciate people nominating BDFL delegates but I don't think it should be accepted without the explicit +1 from me. I've already clarified this in JEP office hours and Tyler.

rtyler:
whoops! :man_facepalming: 

those green buttons are just so fun to click

PR #199, opened 2018-09-12
[WIP] More thoughts about the upgrade and rollback process

batmat:
I'm still writing this as it helps me making sure I dig into details I had previously overlooked. 

So, though this is *not* finished yet, I am eager for feedback here.

### DIFF
--- a/jep/302/README.adoc
+++ b/jep/302/README.adoc
@@ -75,13 +75,54 @@ As Jenkins does not support downgrading alone, this document introduces the snap
 
 _Evergreen_ works with two main components: Jenkins itself, and the _evergreen-client_.
 
-=== Upgrading and downgrading
+=== Safely Upgrading
+
+This section will describe how to _safely_ upgrade:
+i.e. how to apply an update, then restart Jenkins and check it still works, and roll back to the initial state in case it has become _unhealthy_
+footnote:[Cf. <<healthcheck>>.].
+
+==== Challenges and open questions
+
+===== Failed rollback
+
+If an upgrade goes wrong, we will trigger a rollback.
+
+But what if even after that rollback, Jenkins still does not "come back"?
+
+The _current_ design choice is to stop at this point.
+If an instance is unable to brought back to life after a rollback, supposedly using a state where it *was* working, this is unlikely that we'll have a bigger chance to fix it by going back over and over until the _Update Level 0_, with an empty Jenkins Evergreen instance.
+
+It is probably wiser to:
+* report this to the backend, and have the failure analyzed by a human, and then
+* apply the next public _Update Level_ available
+
+===== Infinite upgrade/rollback loop
+
+As per the current design, an instance that is failing is on purpose not authoritative enough to globally _taint_ a given _Update Level_ alone.
+This would be too risky that a single instance can just taint all update levels and stall the whole system.
+
+This means that, after a rollback occurred for an instance to apply that _Update Level X_, that same instance would be presented again and again with that same "bad" _UL X_, and keep upgrading and rolling back forever.
+
+Hence, we need something to _remember a previously failed instance should not retry an Update Level that already caused a rollback.
+Given our current principle of keeping the evergreen client _rather_ simple, doing this on the server is more likely what we should do.
+
+I suggest we create a new endpoint for this purpose: it will look similar to global one allowing to tainting an update level globally (touching the `updates` table).
+The difference being that it will store the couple `(instance uuid|update level)`.
+
+Then, when the client asks for updates, the server will not filter out globally tainted updated, but also the ones that have been known already to make this very instance fail.
+
+From a Evergreen Project operation standpoint, we will obviously need to carefully monitor the calls made to this "instance-by-instance-tainting" endpoint, so that we decide to globally mark a problematic _update level_ as tainted or not.
+
+We could even automate it partly by implementing some rules (to be defined): for instance, if more than 25% of instances are reporting to fail to upgrade to a given _update level_, this update would automatically be flagged preventively as globally tainted to allow humans to analyze the issue more closely before it creates more problems.
+
+
+==== Upgrade Process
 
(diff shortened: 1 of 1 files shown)


jep/302/README.adoc
```diff
@@ -75,13 +75,54 @@ As Jenkins does not support downgrading alone, this document introduces the snap
...
+
+As per the current design, an instance that is failing is on purpose not authoritative enough to globally _taint_ a given _Update Level_ alone.
+This would be too risky that a single instance can just taint all update levels and stall the whole system.
+
+This means that, after a rollback occurred for an instance to apply that _Update Level X_, that same instance would be presented again and again with that same "bad" _UL X_, and keep upgrading and rolling back forever.
+
+Hence, we need something to _remember a previously failed instance should not retry an Update Level that already caused a rollback.
+Given our current principle of keeping the evergreen client _rather_ simple, doing this on the server is more likely what we should do.
+
+I suggest we create a new endpoint for this purpose: it will look similar to global one allowing to tainting an update level globally (touching the `updates` table).
```
rtyler: A JEP is not the place for this kind of suggestion, but rather should prescribe a specific backend design change. 

Though this pull request looks more like a mental scratchpad for @batmat at the moment

jep/302/README.adoc
```diff
@@ -75,13 +75,54 @@ As Jenkins does not support downgrading alone, this document introduces the snap
...
+The difference being that it will store the couple `(instance uuid|update level)`.
+
+Then, when the client asks for updates, the server will not filter out globally tainted updated, but also the ones that have been known already to make this very instance fail.
+
+From a Evergreen Project operation standpoint, we will obviously need to carefully monitor the calls made to this "instance-by-instance-tainting" endpoint, so that we decide to globally mark a problematic _update level_ as tainted or not.
+
+We could even automate it partly by implementing some rules (to be defined): for instance, if more than 25% of instances are reporting to fail to upgrade to a given _update level_, this update would automatically be flagged preventively as globally tainted to allow humans to analyze the issue more closely before it creates more problems.
```
rtyler: This makes sense to me, but the phrasing should become more authoritative if this will make it into the  JEP. There's no "cloud" in JEPs :smile_cat:

jep/302/README.adoc
```diff
@@ -75,13 +75,54 @@ As Jenkins does not support downgrading alone, this document introduces the snap
...
+As per the current design, an instance that is failing is on purpose not authoritative enough to globally _taint_ a given _Update Level_ alone.
+This would be too risky that a single instance can just taint all update levels and stall the whole system.
+
+This means that, after a rollback occurred for an instance to apply that _Update Level X_, that same instance would be presented again and again with that same "bad" _UL X_, and keep upgrading and rolling back forever.
+
+Hence, we need something to _remember a previously failed instance should not retry an Update Level that already caused a rollback.
+Given our current principle of keeping the evergreen client _rather_ simple, doing this on the server is more likely what we should do.
+
+I suggest we create a new endpoint for this purpose: it will look similar to global one allowing to tainting an update level globally (touching the `updates` table).
+The difference being that it will store the couple `(instance uuid|update level)`.
```
rtyler: I think this approach generally makes sense.

If we would want to return the latest valid update level to a client, I think we would increase the complexity in the current update level generation for the client, in that we would basically need to execute a subquery, e.g.

 `WHERE updates.id NOT IN (SELECT tainted.updateId FROM tainted WHERE tainted.uuid = uuid)`

Another route that comes to mind for me: don't tell the client anything if their latest update level is tainted already, for example:

1. Compute the update level for the client
1. Consult the `tainted` table
1. If the update level exist in the `tainted` level, return a 304 Not Modified back to the client.

jep/302/README.adoc
batmat: Yeah, that is right. I was intending to clarify this progressively, hence my 'WIP' title. I'm sorry to use this place, but re-reading, modifying and thinking using the JEP made sense for my mental process.

So, yes, I intend to remove/rephrase this kind of sentence before I declare the PR ready.
Now I would understand we should not use that repo for this, this just looked like the simplest way and logical place to move forward and have that very discussion we're now having.

jep/302/README.adoc
batmat: Will do. I do not think I will want to provide some specific rule already, but saying that "we *will* implement an automated preventive tainting if enough instances report a failure on a given update level"

jep/302/README.adoc
rtyler: :man_shrugging:  it's fine by me to use pull requests her for clarifying thoughts and working it out.

I just wanted to make sure that this got refactored or tidied up before we merge any changes here.

bitwiseman:
@batmat @rtyler 
Is there progress being made here? If not, perhaps it would be better to merge the comments as `WARNING` sections so that those reading the JEP can see what is going on.

batmat:
@bitwiseman TBH, I wrote all this as a way to reflect on the all process more deeply. I would need to re-read what I wrote to be sure about what next step I would recommend.

I am going to have a fresh look at this on next Monday morning to be able to answer here :).

batmat:
Closing this for now, I'll reopen and adjust once I finalize the code associated to this.

bitwiseman:
@batmat  it would be okay to merge this as a reflection of your progress.  I'd rather see this merged with blocks for Tyler's feedback, rather than have this information hidden only in a PR.  (Tyler expressed the same in his comments).

batmat:
@bitwiseman that's fine by me as anyway this JEP is still only in Draft state. I'd just rather not spend too much time on this right now, and for now spend it on finalizing the work to make this actually work. 

So from my PoV we can either merge as-is with some tweaks here and there, or I'll spend time on this a bit later.

batmat:
Unfortunately no time to work on this currently. Will reopen or open a new one when I get back to it. Thanks for the review!

PR #202, opened 2018-09-14
Travel Grant Enhancements JEP

tracymiranda:


daniel-beck:
Updating something else seems inherently prone to problems.

Should rather migrate the existing travel grant program, similar to the JEP about recommended plugins, and then superseded in a new JEP with "Travel Grant Program 2.0".

bitwiseman:
@daniel-beck Please see #206.

PR #206, opened 2018-09-26
Integrate full program into JEP

bitwiseman:
@tracymiranda 
I've updated this JEP with the full text from the wikipage. If we accept this merge, we can point the wikipage to this JEP and use this as the source of truth.

### DIFF
--- a/jep/12/README.adoc
+++ b/jep/12/README.adoc
@@ -37,16 +37,15 @@ endif::[]
 
 == Abstract
 
-The Jenkins project provides travel grants to a limited number of community members requesting assistance to attend event.
+The Jenkins project provides travel grants to a limited number of community members to attend Jenkins related events.
+This document describes the process by which travel grants may be requested and approved.
 
 == Specification
 
-This document describes the process by which travel grants may be requested and approved, and who is eligible to apply.
-
 === Eligibility
 
-All community members are eligible for support unless they've received a travel grant within the last year (based on the event's date).
-However, as we have very limited funds to support this program, we'll prefer applications by active contributors to the Jenkins project.
+All community members are eligible for travel grants.
+However, as we have very limited funds to support this program, each community member may only receive a travel grant once per year (based on the date of the event they plan to attend).
 Grants are offered to individual community members, not companies. 
 
 If you have other possible funding sources, please look to them first.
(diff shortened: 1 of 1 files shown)


jep/12/README.adoc
```diff
@@ -36,24 +36,62 @@ endif::[]
...
 == Specification
 
-[WARNING]
-====
-*(bitwiseman)*
-Rather than this being an update to Travel Grant Policy on the wiki,
-I'd like to see this replace the wiki page entirely.
-====
+The Jenkins project provides travel grants to a limited number of community members requesting assistance.
+Each travel grant supports one individual's travel to a Jenkins community event such as Jenkins User Conferences with up to 500 USD.
```
samrocketman: I understand money is not unlimited but a I have attended the last two Jenkins user conferences and they cost upwards of around $3000 for the event in San Francisco for just flight and hotel.  This grant seems low for that as an example.

bitwiseman: @samrocketman 
I see your point.  At the same time, I don't think the project's budget has ever been sufficient to do more than what is outlined.  

It doesn't sound like your object to the intent of this PR, but want further changes to the program.

Can I ask you to raise this feedback on the discussion thread for this JEP or via a separate PR?

daniel-beck: I think that's addressed by https://github.com/jenkinsci/jep/pull/206/files#diff-e56f06fbe470a002ac6879a018b56756R93

Perhaps some internal inconsistency between this part and the linked?

samrocketman: I think the change @daniel-beck mentions is good.  I mentioned the SF cost mainly because those who would use this kind of budget are likely in need of it.  That is to say they don’t have an alternate sponsor (company) and likely can’t afford to attend otherwise (personal finances, etc).  This was my primary concern with raising the point.

bitwiseman:
@tracymiranda  Do you approve this change?

tracymiranda:
Looks good generally, one comment for first para of spec I suggest we remove bit that says 'up to 500 USD' as that is not strictly true now and we clarify the amounts later- just want to avoid people reaching that bit and thinking that is the only limit.

bitwiseman:
@tracymiranda
Basically rewrote to integrate and clarify.

bitwiseman:
@samrocketman I think I addressed you concern in the reasoning section.

bitwiseman:
@jenkinsci/jep-editors @tracymiranda  This seems generally landed.  Can I get an approval?

PR #207, opened 2018-09-30
removed meta-model proposal, as this collides with data-api

ndeloof:
https://github.com/kohsuke/jep/tree/data-api/jep/0000/1


ndeloof:
@bitwiseman done

PR #208, opened 2018-10-08
Marking JEP-210 as accepted

jglick:
In fact all the code changes have been merged and (where appropriate) released as beta, so I am not sure if Final would be more appropriate, but I am erring on the conservative side here.


rtyler:
As the BDFL delegate, we'll need @carlossg to weigh in here

PR #209, opened 2018-10-16
Hint about "should I write a JEP"

batmat:
Cf. https://groups.google.com/forum/#!msg/jenkinsci-dev/MCN7x4KSWes/AKADOvXBBAAJ :-)


bitwiseman:
@oleg-nenashev  I think your experiences were edge cases.  For example, we really didn't figure out that a JEP wasn't needed for each SIG until we tried it, so that time wasn't wasted. 👍

PR #214, opened 2018-11-12
[JEP-211] - Explicitly add Java 11 Support Team to the JEP

oleg-nenashev:
Maybe we need the same team for jenkins-infra, not sure.

CC @jenkinsci/java11-support


oleg-nenashev:
@jenkinsci/jep-editors

oleg-nenashev:
Merging as JEP sponsor

oleg-nenashev:
> Merging can be performed automatically with 1 approving review.

So, overrides by admins are prohibited here.
@daniel-beck would you be fine if I override it?

PR #216, opened 2018-11-20
[JEP-211] - Update Docker packaging for Master to reflect the current labeling practices

oleg-nenashev:
As discussed with @batmat and @olblak


oleg-nenashev:
Thanks @bitwiseman ! It will help a lot

PR #222, opened 2018-12-08
JEP Draft: Experimental Jenkins organization on DockerHub

oleg-nenashev:
As discussed at the previous SIG meeting, we want to deploy an infrastructure to allow building experimental images on ci.jenkins.io. I have spent some time to draft a JEP which documents the process.

I believe that @olblak could be the best BDFL Delegate for this JEP

CC @jenkinsci/sig-platform

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,208 @@
+= JEP-0000: Experimental Jenkins organization on DockerHub
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
Apparently one review resets the feedback from the GitHub team, even if the review does not set an explicit approval/ request for change. Need somebody to approve it so that I can merge it as a draft.

oleg-nenashev:
For the future maybe we could select a new number prefix for Infrastructure-focused JEPs (especially if IEPs get migrated here).

bitwiseman:
@oleg-nenashev The JEP is relatively new and there aren't a lot of references to this yet.   If it matters to you, let's select a new number.

oleg-nenashev:
There are already references, let's keep it as is. Not a big deal

PR #225, opened 2018-12-10
[JEP-211] - Be explicit about potential security risks in Java 11 preview when running with Java 11

oleg-nenashev:
@jenkinsci/java11-support @daniel-beck @Wadeck

PR #226, opened 2018-12-10
Accept JEP-211 as a BDFL Delegate

oleg-nenashev:
Staging the JEP-211 approval pull request. All comments have been addressed, #225 is the last one to be merged. The deadline for feedback in Dev List / Platform SIG was Monday, so we can go forward with accepting it.

@jenkinsci/java11-support @jenkinsci/jep-editors

### DIFF
--- a/jep/211/README.adoc
+++ b/jep/211/README.adoc
@@ -24,7 +24,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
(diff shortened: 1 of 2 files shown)


bitwiseman:
@oleg-nenashev  See #227 for minor tweaks I'd like added before acceptance.

oleg-nenashev:
Also integrated the last-minute clarification about Java 10 and Java 12+ support. Ready to go, I believe

PR #229, opened 2018-12-13
Org admin approval process details

martinda:
@oleg-nenashev @jeffpearce @lloydchang 

I wanted to add details to the org admin expense approval process in case an org admin is not available or cannot be reached.

I also clarified that expense request approval decisions are made by consensus.

### DIFF
--- a/jep/8/README.adoc
+++ b/jep/8/README.adoc
@@ -122,6 +122,7 @@ payments to external individuals/organizations for promotion.
 
 * All expenses should be pre-approved with
 Jenkins GSoC org admins before the expenses are incurred, and as a pre-condition for payment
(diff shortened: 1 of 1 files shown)


jeffpearce:
I *almost* suggested we make it 1 week instead. Should we shorten it? I don't want people to be blocked on org admins.

oleg-nenashev:
> I almost suggested we make it 1 week instead. Should we shorten it? I don't want people to be blocked on org admins.

In such case I would set a requirement to:

* Requesters to send a request at least 2 days before the Weekly GSoC meeting
* Requesters to join the meeting and discuss it with org admins on the call

oleg-nenashev:
Thanks to all!

PR #230, opened 2018-12-16
[SECURITY-595] Stapler Request Dispatcher Filtering

daniel-beck:
This JEP corresponds to [SECURITY-595 in the 2018-12-05 security advisory](https://jenkins.io/security/advisory/2018-12-05/#SECURITY-595) and was written during development as the reference design for the solution.

As this was necessarily developed in private to prevent exploitation before we could deliver a fix, the JEP is final despite not having gone through the usual process.


oleg-nenashev:
As a JEP editor  I am not sure what is the purpise of a post-factum JEP. It
is not a "proposal", and it is not open for the discussion IIUC.

If it is submitted as JEP, I believe it needs to be explicitly clarified in
Reasoning (which parts are open for discussion and which are not)

daniel-beck:
> what is the purpose of a post-factum JEP

I submitted it because according to this repo's README:

> A JEP provides a concise technical specification of the feature, describes the motivation for the change, and discusses the rationale behind the design.
> 
> JEPs are the primary mechanism for … documenting the design decisions that have gone into Jenkins.

If there's no interest by JEP editors to have this recorded, works for me.

bitwiseman:
@daniel-beck @oleg-nenashev 
It would be completely reasonable for this to be filed as an "Informational" JEP (like [JEP-2](https://github.com/jenkinsci/jep/blob/master/jep/2/README.adoc)).  These do not require consensus and are meant for exactly this type of situation.  The Security team designed and implemented this internally and do not want to expose it until a fix is published. 

@daniel-beck  
I think this is great.  I hope you continue to submit Security-related informational JEPs like this one as a general practice.  We should start a new JEP numbering section (maybe 5xxx to give plenty of space for future security JEPs).

daniel-beck:
@bitwiseman 

> It would be completely reasonable for this to be filed as an "Informational" JEP

I am not sure https://github.com/jenkinsci/jep/tree/master/jep/1#jep-types applies, as it _does_ describe a feature, but I will follow your advice.

WRT next steps, I will change https://github.com/jenkinsci/jep/pull/230/files#diff-b956eee4a4cad96baf0790dddfecba73R29. Anything else?

> We should start a new JEP numbering section

These would simply be technical specifications or designs, so I'm not sure a new section would be warranted -- it's just the process that is necessarily different.

bitwiseman:
@daniel-beck 
I see your point about standards vs informational. This absolutely does describe a feature change, I mentioned informational only as a way of saying it is completely valid to do "post-facto" JEPs.  

I'll go ahead and set this to draft and we can go from there.

PR #231, opened 2018-12-17
Suggest different name for organiztion

bitwiseman:
`Jenkins4eval` has the problem of people potentially
thinking it applies to the non-existant Jenkins v4.0.

I'm not committed to the name `jenkinsfutures`, but I'm a strong -1
for the name `jenkins4eval`.

@oleg-nenashev

### DIFF
--- a/jep/217/README.adoc
+++ b/jep/217/README.adoc
@@ -64,7 +64,7 @@ endif::[]
 
 == Abstract
 
-This JEP documents the `jenkins4eval` organization on DockerHub.
+This JEP documents the `jenkinsfutures` organization on DockerHub.
 The main purpose of this organization is to provide a storage for experimental Jenkins Docker images
 so that maintainers can build and deploy untrusted images from ci.jenkins.io.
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Please comment in the suggested communication channel

DockerHub does not support org renaming, so it may require migration of all
existing images and docs referencing them. Not a big deal, but this is not
just a rename. Let's revisit it at the next SIG meeting

jenkins4eval as a name and "Jenkins 4" concern were discussed at platform
SIG meetings in Aug IIRC

bitwiseman:
The reasoning section doesn't summarize or link to the discussion of those concerns.  If this was discussed and resolved already, I shouldn't have to go digging through meeting notes to find it.  This is one the purposes of the reasoning section.

oleg-nenashev:
> The reasoning section doesn't summarize or link to the discussion of those concerns. If this was discussed and resolved already, I shouldn't have to go digging through meeting notes to find it.

That's totally right, you are not expected to be aware of it since it was not referenced. And there is no intent to reject your comment by referencing a discussion which happened months ago when this JEP was not even planned.

This JEP (created 1 week ago) proposes to use the organization which de-facto existed by now. It is not what the JEP text suggests, so it needs to be updated anyway. Meeting notes are [here](https://docs.google.com/document/d/1YofL2uhy7xAa1mx_qFdDvDg4P-molmhDwFD0-8xX8mI/edit#heading=h.q612ds3xp868) BTW.

I do not want to update the JEP right now, because I think that we should revisit that decision anyway. So there will be a discussion which will be referenced from the

bitwiseman:
If we're not going to update the JEP with a new name now but we plan to do so, then until the changes are landed let's can add a comment to the effect that the name is being reconsidered.

oleg-nenashev:
works for me

oleg-nenashev:
Clarification has been added in https://github.com/jenkinsci/jep/pull/233

PR #237, opened 2019-01-30
Update JEP number in intro

jvz:
@jenkinsci/jep-editors


oleg-nenashev:
also needs metadata update though, see the script in the root

oleg-nenashev:
nws, I will fix it later

PR #243, opened 2019-05-10
Update status of localization

LinuxSuRen:
localization-support-plugin had released


LinuxSuRen:
@bitwiseman @daniel-beck Please take a look.

PR #244, opened 2019-06-13
Draft: Opt-In Continuous Delivery of Jenkins Plugin Releases

kshultzCB:
This proposal, once complete, will describe a mechanism by which Jenkins plugin maintainers could choose to have their plugins released upon merging to a particular branch. This merge operation would trigger a webhook, which would in turn trigger a build/test/deploy cycle from a trusted CI server managed by the Jenkins infrastructure team.

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -115,14 +115,22 @@ community at large.
 
 === Setup
 
-Two main things:
+==== The autorelease Maven profile
 
-A new Maven profile will be created, most likely in the 
+I actually don't know that this is necessary. A new Maven profile can be created, 
+most likely in the 
 link:https://github.com/jenkinsci/plugin-pom[Jenkins plugin parent POM]. 
 
-* enable-autorelease
+==== Changes to `buildPlugin`
 
-A YAML markerfile, stored in the plugin's own repository. Absence of this markerfile will In the following example, we'll use a fictional plugin, called `fancy-branch-source`
+The commonly used library link:https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy[`buildPlugin`] will need to be modified to check for, and validate, a markerfile, called `Autorelease.yaml`.
+
+==== Autorelease.yaml
+
+A YAML markerfile, stored in the plugin's own repository. Absence of this markerfile will 
+be functionally equivalent to the plugin maintainer opting out of Autorelease. 
+
+In the following example, we'll use a fictional plugin, called `fancy-branch-source`
 
 ```
 plugin: fancy-branch-source
@@ -133,7 +141,7 @@ metadata:
   appendMergeSHA: true
 ```
 
-In the above example, we can see the following:
+In the above example `Autorelease.yaml`, we can see the following:
 
 `plugin: fancy-branch-source`:: 
 This is the name of the plugin, and it matches the repository 
@@ -194,7 +202,7 @@ look like:
         <jenkins.version>2.138.4</jenkins.version>
     </properties>
 
-````
+```
 
 This is all pretty run-of-the-mill stuff for a Jenkins plugin, and is well understood already 
 by Karl and the rest of the plugin maintainer community. To enable autorelease, Karl would 
@@ -211,6 +219,7 @@ metadata:
 
 When Karl merges a commit into the `release-the-fanciness` branch, that merge commit has the 
 SHA `1a2b3c4`. The following takes place:
+
 * A webhook is sent to super-trusted-thing, and a build is performed there. 
 * If the build passes all its tests, a release is generated. In our example, 
 that release number would be `3.1.4-1a2b3c4`, because Karl has chosen to append the merge commit 
@@ -226,47 +235,20 @@ ability to do continuous, merge-driven releases, plugin maintainers can readily 
 Jenkins itself is being worked on in a continuous way. This would be a big win for Jenkins' 
 credibility in an increasingly demanding market space.
 
-[TIP]
-====
-Explain why the existing code base or process is inadequate to address the problem that the JEP solves.
-This section may also contain any historical context such as how things were done before this proposal.
-
-* Do not discuss design choices or alternative designs that were rejected - those belong in the Reasoning section.
-====
-
 == Reasoning
 
-Oh boy the UX for this could really suck. People are going to be overrun with upgrades they think they 
-need to do.
+=== Why Maven?
+With Incrementals having gained fairly wide adoption by plugin maintainers, it seems reasonable to use maven 
 
-Maybe we need an "urgent" flag for those?
 
-[TIP]
-====
-Explain why particular design decisions were made.
-Describe alternate designs that were considered and related work. For example, how the feature is supported in other systems.
-Provide evidence of consensus within the community and discuss important objections or concerns raised during discussion.
-
-* Use sub-headings to organize this section for ease of readability.
-* Do not talk about history or why this needs to be done - that is part of Motivation section.
-====
+=== Why a trusted CI system?
+Containment of credentials. By using a single system of record for these builds, a service account, maintained by the JENKINS-CERT team, can be used to access GitHub, deploy to Nexus, and deploy to the update centers.
 
 == Backwards Compatibility
-
-[TIP]
-====
-Describe any incompatibilities and their severity.
-Describe how the JEP proposes to deal with these incompatibilities.
-
-If there are no backwards compatibility concerns, this section may simply say:
-There are no backwards compatibility concerns related to this proposal.
-====
 
 With any plugin upgrade, there are backwards compatibility concerns, and autorelease is no different 
 in that regard. In fact, it could be made worse, because users will get overrun with upgrades. Jenkins 
 existing plugin system only allows for uninstalling one revision. 
-
-
 
 == Security
 
@@ -293,24 +275,19 @@ this in any way.
 
 == Prototype Implementation
 
-[TIP]
-====
-Link to any open source reference implementation of code changes for this proposal.
-The implementation need not be completed before the JEP is
-link:https://github.com/jenkinsci/jep/tree/master/jep/1#accepted[accepted],
-but must be completed before any JEP is given
+As a proof of concept, the (github-branch-source?) plugin will be the first to adopt. This provides 
+the initiative with a heavily used plugin, which sees relatively frequent releases already.
+
+A sample fork of github-branch-source could be provided as a reference implementation for 
+this proposal. It is understood that this need not be completed before this JEP is 
+"link:https://github.com/jenkinsci/jep/tree/master/jep/1#accepted[accepted]", but will need to 
+be made available before this JEP is given 
 "link:https://github.com/jenkinsci/jep/tree/master/jep/1#final[Final]" status.
 
-JEPs which will not include code changes may omit this section.
-====
 
 == References
 
-[TIP]
-====
-Provide links to any related documents.
-This will include links to discussions on the mailing list, pull requests, and meeting notes.
-====
+None yet.
 
 
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -64,21 +64,13 @@ endif::[]
 
 == Abstract
 
-At present, Jenkins plugins are, typically, not released on a continous basis. They 
-are also not released from a single source of truth, such as a trusted Continuous 
-Integration server like link:https://ci.jenkins.io[https://ci.jenkins.io]. 
+This JEP introduces a new system, which enables Jenkins plugin maintainers to add Continuous 
+Delivery to their projects. It is:
 
-The notion of continuous delivery of plugin releases has been discussed previously <<footnote-1,^(1)^>>.
-Considering that Jenkins is a system used to facilitate Continuous Delivery for many users, it makes 
-sense -- and builds credibility -- for the Jenkins developer community to adopt this same practice. 
-
-Having a centralized release system made available to plugin maintainers also provides additional 
-confidence that security best practices are being followed *need footnote without surfacing some 
-awful security problem* 
-
-Continuous celivery from trusted CI is something which plugin maintainers can opt into, but is 
-not required. If a plugin maintainer chooses to continue to follow their own path for releasing 
-versions of their code, they remain free to do so.
+* Opt-in: plugin maintainers can choose to use this
+* Simple to configure: very little needs to be changed in order to enable CD
+* Trusted CI/CD: deployments are produced from a single source of truth, which is managed by 
+the Jenkins project
 
 == Specification
 
@@ -209,11 +201,21 @@ SHA to the end of his autorelease version numbers.
 
 == Motivation
 
-It's no secret that the Jenkins plugin ecosystem is complex. It's also no secret that Jenkins plugins 
-are often developed in a very non-continuous way. This proposal seeks to change this. By offering the 
-ability to do continuous, merge-driven releases, plugin maintainers can readily make the claim that 
-Jenkins itself is being worked on in a continuous way. This would be a big win for Jenkins' 
-credibility in an increasingly demanding market space.
+At present, Jenkins plugins are, typically, not released on a continous basis. They 
+are also not released from a single source of truth, such as a trusted Continuous 
+Integration server like link:https://ci.jenkins.io[https://ci.jenkins.io]. 
+
+The notion of continuous delivery of plugin releases has been discussed previously <<footnote-1,^(1)^>>.
+Considering that Jenkins is a system used to facilitate Continuous Delivery for many users, it makes 
+sense -- and builds credibility -- for the Jenkins developer community to adopt this same practice. 
+
+Having a centralized release system made available to plugin maintainers also provides additional 
+confidence that security best practices are being followed *need footnote without surfacing some 
+awful security problem* 
+
+Continuous celivery from trusted CI is something which plugin maintainers can opt into, but is 
+not required. If a plugin maintainer chooses to continue to follow their own path for releasing 
+versions of their code, they remain free to do so.
 
 == Reasoning
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -93,23 +93,25 @@ ecosystem, Karl settles on link:https://semver.org/[semantic versioning].
 
 4. Add a file to the top directory of his repository called `Autorelease.yaml`. 
 
-5. A webhook is configured on Karl's repository on GitHub to point to `{the-webhook-receiver}`
+5. A webhook is configured on Karl's repository on GitHub to point to a webhook receiver. 
 
 Once ready to create a new release of his plugin, Karl merges some changes to the `master` branch. 
 At which point, the following takes place under the covers: 
 
-1. A webhook is sent to `{the-webhook-receiver}` by GitHub
+1. A webhook is sent to the webhook receier by GitHub
 
-2. The receiver forwards this hook to `${trusted-ci-server}`, which is protected by a VPN
+2. The webhook receiver forwards this hook to a trusted CI server, which is protected by a VPN
 
-3. `${trusted-ci-server}` uses the enhanced version of `buildPlugin` to validate the correctness of 
-Karl's `Autorelease.yaml` file, and the consistency of Karl's chosen version numbering.
+3. The trusted CI server, upon getting the webhook from the trusted receiver, uses the 
+enhanced version of `buildPlugin` to validate the correctness of Karl's `Autorelease.yaml` 
+file, and the consistency of Karl's chosen version numbering.
 
-4. Once `Autorelease.yaml` has been validated, a plugin build is performed on `${trusted-ci-server}` 
-as one would expect.
+4. Once `Autorelease.yaml` has been validated, a plugin build is performed on the trusted CI server.
 
-5. After the build is successful, `${trusted-ci-server}` handles deployment of the built plugin to 
-Artifactory and the Jenkins Update Center.
+5. After the build is successful, the trusted CI server handles deployment of the built plugin to
+Artifactory, at which point 
+link:https://github.com/jenkins-infra/update-center2/blob/master/README.md[the existing UC generator] 
+makes the plugin available on the update center.
 
 === Detailed Setup
 
@@ -121,34 +123,26 @@ be functionally equivalent to the plugin maintainer opting out of Autorelease.
 In the following example, we'll stick with plugin developer Karl, and his fictional 
 plugin `karl-plugin`. Karl's `Autorelease.yaml` file looks like this:
 
-```
-plugin: karl
-kind: plugin
-metadata:
-  autoreleaseEnabled: true
-  triggeringBranch: master
-  versionNumberFields: 3
-```
+[source,yaml]
+----
+kind: plugin <1>
+triggeringBranch: master <2>
+versionNumberFields: 3 <3>
+----
 
-In the above example `Autorelease.yaml`, we can see the following:
-
-`plugin: karl`:: 
-This is the name of the plugin, and it matches the repository 
-name on GitHub.
-`kind`:: 
+<1> `kind`
 The type of project which is being set up for autorelease. In our first iteration of 
 autorelease, this will only be set to `plugin`. The field is here should the Jenkins project 
 want to expand the use of autorelease, such as for building core, or components.
-`metadata`:: 
-Contains the following fields:
-** `autoreleaseEnabled`: A boolean value, indicating whether or not this plugin is participating 
-in autorelease. Allows the maintainer to turn Autorelease on or off, without deleting the file 
-entirely.
-** `triggeringBranch`: Sets the branch which maintainers will merge to in order to trigger 
+
+<2> `triggeringBranch`
+Sets the branch which maintainers will merge to in order to trigger 
 an autorelease webhook. Although this is expected to usually be set to `master`, plugin 
 maintainers can choose a branch name of their preference, e.g., `autorelease`, `release`, 
 etc.
-** `versionNumberFields`: An integer, indicating the number of fields which `buildPlugin` should expect.
+
+<3> `versionNumberFields`
+An integer, indicating the number of fields which `buildPlugin` should expect.
 This is important, because sticking to a consistent numbering scheme makes for easier sorting by version.
 For this example plugin, Karl has set the value to `3`, because he's usuing semantic versioning, e.g., 
 3.1.4.
@@ -312,7 +306,7 @@ We will need a number of things to get this going. The low-level technical detai
 in a separate Infrastructure Enhancement Request, so, this should be considered a summary for now:
 
 1. The webhook, configured on each participating plugin. Security implications of this are a bit beyond the scope of this document so far.
-2. A receiver for the aforementioned webhook, because the `${trusted-ci-server}` will be protected 
+2. [[webhook-receiver]]A receiver for the aforementioned webhook, because the `${trusted-ci-server}` will be protected 
 behind a VPN
 3. A trusted Jenkins server for performing builds and deployments. 
 4. `buildPlugin` will need code added to validate `Autorelease.yaml` for correctness, and 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -191,9 +191,7 @@ look like:
 
 No additional changes need to be made to `pom.xml` by Karl, he's good to go.
 
-=== What happens when a release is created
-
-==== Changes to `buildPlugin`
+==== Autorelease.yaml verification in `buildPlugin`
 
 The commonly used library link:https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy[`buildPlugin`] will need to be modified to check for, and validate, the markerfile, called `Autorelease.yaml`. This 
 markerfile will be the mechanism that tells `${trusted-ci-server}` that this plugin should be automatically 
@@ -201,20 +199,26 @@ released.
 
 Validation will include:
 
-1. The 
+1. `pluginName` field must match the repository name
+
+2. `versionNumberFields` must match the `revision` property in `pom.xml`. E.g., if `versionNumberFields` is 
+set to `3`, and the `revision` is set to `3.1.4.5`, validation will fail.
+
+3. Validation that no extra fields are present in the file. Comments are allowed, but any unexpected 
+lines will cause validation to fail.
 
 If validation of `Autorelease.yaml` fails for any reason, the build is not performed and nothing gets deployed.
 
+==== Deployment of a successful release
 
-When Karl merges a commit into the `master` branch, that merge commit has the 
+When Karl merges a commit into the `master` branch, that merge commit is commit number 150, and has the 
 SHA `1a2b3c4`. The following takes place:
 
-* A webhook is sent to `${trusted-ci-server}`, and a build is performed there. 
-* If the build passes all its tests, a release is generated. In our example, 
-that release number would be `3.1.4-1a2b3c4`, because Karl has chosen to append the merge commit 
-SHA to the end of his autorelease version numbers.
+* A webhook is sent to `${trusted-ci-server}`, and `Autorelease.yaml` is validated.
+* Once validation passes, a build is performed. If the build passes all its tests, a release 
+is generated. In our example, that release number would be `3.1.4-150-1a2b3c4`
 * The built plugin gets deployed to Nexus
-* The resulting plugin appears on the Jenkins Update Center
+* The resulting plugin appears on the Jenkins Update Center, as Fancy Branch Plugin version 3.1.4-150-1a2b3c
 
 == Motivation
 
@@ -236,8 +240,17 @@ versions of their code, they remain free to do so.
 
 == Reasoning
 
-=== Why a trusted CI system?
-Containment of credentials. By using a single system of record for these builds, a service account, maintained by the JENKINS-CERT team, can be used to access GitHub, deploy to Nexus, and deploy to the update centers.
+=== What's with the version numbering?
+
+Jenkins plugin maintainers are already familiar with the way that Incrementals appends a commit 
+number, pluse a merge SHA, to version numbers. These mechanically-generated version numbers offer 
+the ability to predictably sort them, so that external systems, such as the Jenkins update center, 
+can correctly publish the "newest" version. The addition of a merge SHA also allows for at-a-glance 
+feedback to tell people which commits went into the release.
+
+NOTE: This version number management scheme is only one of several possible choices. We expect 
+some lively debate around this topic. But it's very important to ensure that versions can be 
+easily sorted by systems such as the Jenkins update center.
 
 === Why YAML?
 YAML is becoming increasingly common in the Jenkins community, for many reasons. YAML is:
@@ -247,11 +260,18 @@ YAML is becoming increasingly common in the Jenkins community, for many reasons.
 * In use for things like the Kubernetes plugin
 * Easily parsed by any number of publily available libraries
 
+=== Aren't we missing some details about infrastructrue requirements?
+
+In short, yes, we are. This JEP exists to get the conversation started. Once consensus has been reached, 
+a separate Infrastructure Enhancement Proposal (IEP) will be created to go along with this JEP. It would 
+be premature to describe every detail of implementation before consensus is reached.
+
 == Backwards Compatibility
 
 Autorelease introduces no new risks with regard to backwards compatibility of plugins themselves. 
 However, there will be a requirement for plugin maintainers to use a consistent version numbering 
-scheme. Consistent version numbering will prevent problems with sorting versions 
+scheme. Consistent version numbering will prevent problems with sorting versions. This is discussed 
+elsewhere in this document.
 
 With any plugin upgrade, there are backwards compatibility concerns, and Autorelease is no different 
 in that regard. Without Autorelease, there is still nothing stopping a plugin maintainer from releasing 
@@ -262,15 +282,25 @@ maintainers will see no change at all to the way they do their Jenkins plugin wo
 
 == Security
 
-Autorelease should make things more secure, because they all come from `${trusted-ci-server}`. Rules 
+Autorelease can make things more secure
+
+=== Containment of credentials
+
+By using a single system of record for these builds, a service account, maintained by the 
+JENKINS-CERT team, can be used to access GitHub, deployment to Nexus, and deployment to 
+the update centers.
+
+=== Automatic enforcement of security best practices
+Autorelease builds will all come from `${trusted-ci-server}`. Rules 
 can be put in place on `${trusted-ci-server}`, which can provide implicit enforcement of the 
-Jenkins infrastructure team's security best practices.
+Jenkins infrastructure team's security best practices. Compliance to these best practices becomes 
+something that plugin maintainers need not worry about.
 
 == Infrastructure Requirements
 
 We will need:
 
-1. The webhook. Security implications of this are a bit beyond the scope of this document so far.
+1. The webhook, configured on each participating plugin . Security implications of this are a bit beyond the scope of this document so far.
 ALSO a receiver for the webhook. This 
 2. A receiver for the aforementioned webhook, because the `${trusted-ci-server}` will be protected 
 behind a VPN
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -205,7 +205,7 @@ will fail.
 3. Validation that no extra fields are present in the file. Comments are allowed, but any unexpected 
 lines will cause validation to fail.
 
-If validation of `Autorelease.yaml` fails for any reason, the build is not performed nothing gets deployed, 
+If validation of `Autorelease.yaml` fails for any reason, the build is not performed, nothing gets deployed, 
 and GitHub is notified of the failure.
 
 ==== Deployment of a successful release
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -97,7 +97,7 @@ as that is the simplest.
 Once ready to create a new release of his plugin, Karl merges some changes to the `master` branch. 
 At which point, the following takes place under the covers: 
 
-1. A webhook is sent to the webhook receiver by GitHub
+1. A webhook is sent to a webhook receiver by GitHub
 
 2. The webhook receiver forwards this hook to a trusted CI server, which is protected by a VPN
 
@@ -262,6 +262,28 @@ YAML is becoming increasingly common in the Jenkins community, for many reasons.
 * Human readable
 * In use by things like the Kubernetes plugin
 * Easily parsed by any number of publicly available libraries
+
+=== Webhooks and the Webhook Receiver
+Webhooks will be used to trigger the builds on the trusted CI server. The hooks themselves 
+will be configured on a per-repository basis. The reasons for choosing individual hooks, 
+as opposed to an organization wide hook, are:
+
+* Security of the build systems. Because this trusted CI server will reside on a non-public 
+network, there needs to be a proxy server of sorts, which can accept these webhooks and send 
+them, securely, to the trusted CI server.
+
+
+* Traceability. An organization-wide webhook would send those hooks to the receiver far more 
+often than is actually necessary. In the event that something goes wrong with this process - 
+the receiver goes down, there is a network outage, etc. - it will be easier for the Infra 
+team to triage problems if they're only looking at hooks from plugins which are intended 
+to be released via CD.
+
+* Flexibility for plugin maintainers. Let's say that Karl maintains two plugins - his archaic 
+`old-timey-plugin`, and his fancy new `karl-plugin` which we've already talked about. If 
+Karl wants to use CD on karl-plugin, but continue to release `old-timey-plugin` from his laptop 
+like he's always done, he can do so, by only setting the webhook up on `karl-plugin`.
+
 
 === Testing Considerations
 Continuous Delivery brings with it a heightened importance for quality automated tests. However, 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -229,7 +229,6 @@ YAML is becoming increasingly common in the Jenkins community, for many reasons.
 * Human readable
 * In use for things like the Kubernetes plugin
 * Easily parsed by any number of publily available libraries
-* 
 
 == Backwards Compatibility
 
@@ -246,13 +245,13 @@ maintainers will see no change at all to the way they do their Jenkins plugin wo
 
 Autorelease should make things more secure, because they all come from `${trusted-ci-server}`. Rules 
 can be put in place on `${trusted-ci-server}` which prevent people from doing silly things.
-It also eliminates the potential for MitM attacks.
 
 == Infrastructure Requirements
 
 We will need:
 
 1. The webhook. Security implications of this are a bit beyond the scope of this document so far.
+ALSO a receiver for the webhook. This 
 2. The trusted Jenkins server doing these builds will need to be smart enough to understand the 
 `Autorelease.yaml` file, and act according to its settings 
 3. Probably `buildPlugin` will need some code added to validate the contents of `Autorelease.yaml`. 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -92,6 +92,8 @@ as that is the simplest.
 
 4. Add a file to the top directory of his repository called `Autorelease.yaml`. 
 
+WARNING: *jvz* Couldn't this be done as an argument to buildPlugin()?
+
 5. A webhook is configured on Karl's repository on GitHub to point to a webhook receiver. 
 
 Once ready to create a new release of his plugin, Karl merges some changes to the `master` branch. 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -220,15 +220,29 @@ credibility in an increasingly demanding market space.
 === Why a trusted CI system?
 Containment of credentials. By using a single system of record for these builds, a service account, maintained by the JENKINS-CERT team, can be used to access GitHub, deploy to Nexus, and deploy to the update centers.
 
+=== Why YAML?
+YAML is becoming increasingly common in the Jenkins community, for many reasons. YAML is:
+
+* Already in use by the Tekton project in Jenkins-X
+* Human readable
+* In use for things like the Kubernetes plugin
+* Easily parsed by any number of publily available libraries
+* 
+
 == Backwards Compatibility
+
+Autorelease introduces no new risks with regard to backwards compatibility or a lack thereof.
 
 With any plugin upgrade, there are backwards compatibility concerns, and Autorelease is no different 
 in that regard. Without Autorelease, there is still nothing stopping a plugin maintainer from releasing 
 a backwards-breaking change.
 
+Plugin maintainers are also not required to use Autorelease at all. By taking no action, these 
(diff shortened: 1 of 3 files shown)


jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+4. Add a file to the top directory of his repository called `Autorelease.yaml`. 
+
+5. A webhook is configured on Karl's repository on GitHub to point to `{the-webhook-receiver}`
+
+Once ready to create a new release of his plugin, Karl merges some changes to the `master` branch. 
+At which point, the following takes place under the covers: 
+
+1. A webhook is sent to `{the-webhook-receiver}` by GitHub
+
+2. The receiver forwards this hook to `${trusted-ci-server}`, which is protected by a VPN
```
jglick: Again, `${trusted-ci-server}` is confusing syntax. Is this supposed to be an actual code variable? Or do you just mean to say

> a trusted CI server

where the details will be spelled out in another section?

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+3. `${trusted-ci-server}` uses the enhanced version of `buildPlugin` to validate the correctness of 
+Karl's `Autorelease.yaml` file, and the consistency of Karl's chosen version numbering.
+
+4. Once `Autorelease.yaml` has been validated, a plugin build is performed on `${trusted-ci-server}` 
+as one would expect.
+
+5. After the build is successful, `${trusted-ci-server}` handles deployment of the built plugin to 
+Artifactory and the Jenkins Update Center.
```
jglick: To both? Or to Artifactory, which then causes the plugin to appear on the UC by virtue of the existing UC generator?

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+```
+plugin: karl
+kind: plugin
+metadata:
+  autoreleaseEnabled: true
+  triggeringBranch: master
+  versionNumberFields: 3
+```
+
+In the above example `Autorelease.yaml`, we can see the following:
```
jglick: [Callouts](https://www.methods.co.nz/asciidoc/chunked/ch20.html) would help here.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+`plugin: karl`:: 
+This is the name of the plugin, and it matches the repository 
+name on GitHub.
```
jglick: Well, no, it matches the `artifactId` I suppose, where the repository name is `${artifactId}-plugin`.

And in such a case, why specify it at all? At best it is going to be a source of typos and mistakes; at worst a security risk.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+`kind`:: 
+The type of project which is being set up for autorelease. In our first iteration of 
+autorelease, this will only be set to `plugin`. The field is here should the Jenkins project 
+want to expand the use of autorelease, such as for building core, or components.
+`metadata`:: 
+Contains the following fields:
+** `autoreleaseEnabled`: A boolean value, indicating whether or not this plugin is participating 
+in autorelease. Allows the maintainer to turn Autorelease on or off, without deleting the file
```
jglick: IMHO drop this field. We can comfortably delete files, because we have version control.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+In the above example `Autorelease.yaml`, we can see the following:
+
+`plugin: karl`:: 
+This is the name of the plugin, and it matches the repository 
+name on GitHub.
+`kind`:: 
+The type of project which is being set up for autorelease. In our first iteration of 
+autorelease, this will only be set to `plugin`. The field is here should the Jenkins project 
+want to expand the use of autorelease, such as for building core, or components.
+`metadata`::
```
jglick: Why bother with the subsection? This is all “metadata”.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+markerfile will be the mechanism that tells `${trusted-ci-server}` that this plugin should be automatically 
+released. 
+
+Validation must include, but will not be limited to, the following:
+
+1. `pluginName` field must match the repository name
+
+2. `versionNumberFields` must match the `revision` property in `pom.xml`. E.g., if `versionNumberFields` is 
+set to `3` (three fields), and the `revision` in `pom.xml` is set to `3.1.4.5` (four fields), validation 
+will fail.
```
jglick: So…why require it to be specified to begin with?

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+This is important, because sticking to a consistent numbering scheme makes for easier sorting by version.
+For this example plugin, Karl has set the value to `3`, because he's usuing semantic versioning, e.g., 
+3.1.4.
+
+==== Maven POM file changes
+
+Karl's plugin is already making use of Incrementals, and the relevant lines of his `pom.xml` file 
+look like:
+
+```
```
jglick:
```suggestion
[source,xml]
----
```

for highlighting (terminate with another `----`).

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+maintainers can choose a branch name of their preference, e.g., `autorelease`, `release`, 
+etc.
+** `versionNumberFields`: An integer, indicating the number of fields which `buildPlugin` should expect.
+This is important, because sticking to a consistent numbering scheme makes for easier sorting by version.
+For this example plugin, Karl has set the value to `3`, because he's usuing semantic versioning, e.g., 
+3.1.4.
+
+==== Maven POM file changes
+
+Karl's plugin is already making use of Incrementals, and the relevant lines of his `pom.xml` file
```
jglick:
```suggestion
Karl's plugin is already making use of link:https://github.com/jenkinsci/incrementals-tools/blob/master/README.md#usage-in-plugin-poms[Incrementals], and the relevant lines of his `pom.xml` file 
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+3. Validation that no extra fields are present in the file. Comments are allowed, but any unexpected 
+lines will cause validation to fail.
+
+If validation of `Autorelease.yaml` fails for any reason, the build is not performed nothing gets deployed, 
+and GitHub is notified of the failure.
+
+==== Deployment of a successful release
+
+When Karl merges a commit into the `master` branch, that merge commit is commit number 150, and has the
```
jglick:
```suggestion
When Karl merges a commit into the `master` branch, that merge commit is link:https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc#basic-usage[commit number] 150, and has the 
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+If validation of `Autorelease.yaml` fails for any reason, the build is not performed nothing gets deployed, 
+and GitHub is notified of the failure.
+
+==== Deployment of a successful release
+
+When Karl merges a commit into the `master` branch, that merge commit is commit number 150, and has the 
+SHA `1a2b3c4`. The following takes place:
+
+* A webhook is sent to `${trusted-ci-server}`, and `Autorelease.yaml` is validated.
+* Once validation passes, a build is performed. If the build passes all its tests, a release
```
jglick: This build is happening on trusted.ci.jenkins.io or whatever, in addition to the build happening on ci.jenkins.io? Or is ci.jenkins.io the “trusted CI server”?

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+==== Deployment of a successful release
+
+When Karl merges a commit into the `master` branch, that merge commit is commit number 150, and has the 
+SHA `1a2b3c4`. The following takes place:
+
+* A webhook is sent to `${trusted-ci-server}`, and `Autorelease.yaml` is validated.
+* Once validation passes, a build is performed. If the build passes all its tests, a release 
+is generated. In our example, that release number would be `3.1.4-150-1a2b3c4`
+* The built plugin gets deployed to Artifactory
+* The resulting plugin appears on the Jenkins Update Center, as Karl Plugin version `3.1.4-150-1a2b3c4`.
```
jglick: Appears how?

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+==== Deployment of a successful release
+
+When Karl merges a commit into the `master` branch, that merge commit is commit number 150, and has the 
+SHA `1a2b3c4`. The following takes place:
+
+* A webhook is sent to `${trusted-ci-server}`, and `Autorelease.yaml` is validated.
+* Once validation passes, a build is performed. If the build passes all its tests, a release 
+is generated. In our example, that release number would be `3.1.4-150-1a2b3c4`
+* The built plugin gets deployed to Artifactory
```
jglick: This already happens in ci.jenkins.io builds. The only thing that does _not_ happen today is that the update center does not pick up incremental builds (from `master`).

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+A sample fork of github-branch-source could be provided as a reference implementation for 
+this proposal. It is understood that this need not be completed before this JEP is 
+"link:https://github.com/jenkinsci/jep/tree/master/jep/1#accepted[accepted]", but will need to 
+be made available before this JEP is given 
+"link:https://github.com/jenkinsci/jep/tree/master/jep/1#final[Final]" status.
+
+
+== References
+
+[[footnote-1]]1. Jenkins World 2017, link:http://bit.ly/2x1lCUZ[Contributor Summit Notes], pp. 11-12
```
jglick:
```suggestion
[[footnote-1]]1. Jenkins World 2017, link:[Contributor Summit Notes]
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+Integration server like link:https://ci.jenkins.io[https://ci.jenkins.io]. 
+
+The notion of continuous delivery of plugin releases has been discussed previously <<footnote-1,^(1)^>>.
+Considering that Jenkins is a system used to facilitate Continuous Delivery for many users, it makes 
+sense -- and builds credibility -- for the Jenkins developer community to adopt this same practice. 
+
+Having a centralized release system made available to plugin maintainers also provides additional 
+confidence that security best practices are being followed. 
+
+Continuous delivery from trusted CI is something which plugin maintainers can opt into, but is
```
jglick:
```suggestion
Continuous delivery from trusted CI is something which plugin maintainers can opt in to, but is 
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+that versions can be easily sorted by systems such as the Jenkins update center. Also under 
+consideration is defining a fixed number of fields in version numbers, and considering any 
+deviation from that a validation failure.
+
+=== Why YAML?
+YAML is becoming increasingly common in the Jenkins community, for many reasons. YAML is:
+
+* Human readable
+* In use by things like the Kubernetes plugin
+* Easily parsed by any number of publily available libraries
```
jglick:
```suggestion
* Easily parsed by any number of publicly available libraries
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+feedback to tell people which commits went into the release.
+
+WARNING: We know the problem of consistent version numbering, and the resulting ability to 
+sort versions, is important. The scheme discussed here is only one of several possible 
+choices. We expect some _lively debate_ around this topic. But it's very important to ensure 
+that versions can be easily sorted by systems such as the Jenkins update center. Also under 
+consideration is defining a fixed number of fields in version numbers, and considering any 
+deviation from that a validation failure.
+
+=== Why YAML?
```
jglick: Can probably be dropped. I cannot imagine anyone raising an eyebrow over _this_ aspect.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+=== Aren't we missing some details about infrastructrue requirements?
+
+In short, yes, we are. This JEP exists to get the conversation started. Once consensus has been reached, 
+a separate Infrastructure Enhancement Proposal (IEP) will be created to go along with this JEP. It would 
+be premature to describe every detail of implementation before consensus is reached.
+
+== Backwards Compatibility
+
+Autorelease introduces no new risks with regard to backwards compatibility of plugins themselves.
```
jglick: Except to the extent that plugin maintainers might forget that merging a PR constitutes public release—there is no separate stage of manually validating behavior and thinking about changes before cutting a release.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+in that regard. Without Autorelease, there is still nothing stopping a plugin maintainer from releasing 
+a backwards-breaking change.
+
+Plugin maintainers are also not required to use Autorelease at all. By taking no action, these 
+maintainers will see no change at all to the way they do their Jenkins plugin work.
+
+== Security
+
+Autorelease can help to make Jenkins plugins, and their release processes, more secure in a 
+number of ways. Including, but not limited to:
```
jglick:
```suggestion
number of ways, including but not limited to:
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+=== Containment of credentials
+
+By using a single system of record for these builds, a service account, maintained by the 
+JENKINS-CERT team, can be used to access GitHub, deployment to Nexus, and deployment to
```
jglick:
```suggestion
Jenkins CERT team, can be used to access GitHub, deployment to Nexus, and deployment to 
```

jglick:
```suggestion
JENKINS-CERT team, can be used to access GitHub, deploy to Nexus, and deploy to 
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+== Security
+
+Autorelease can help to make Jenkins plugins, and their release processes, more secure in a 
+number of ways. Including, but not limited to:
+
+=== Containment of credentials
+
+By using a single system of record for these builds, a service account, maintained by the 
+JENKINS-CERT team, can be used to access GitHub, deployment to Nexus, and deployment to 
+the update centers. Plugin maintainers need not leave their own credentials on a CI server
```
jglick: In fact they need not have Artifactory credentials at all.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+=== Automatic enforcement of security best practices
+Autorelease builds will all come from `${trusted-ci-server}`. Rules 
+can be put in place on `${trusted-ci-server}`, which can provide implicit enforcement of the 
+Jenkins infrastructure team's security best practices. Compliance to these best practices becomes 
+something that plugin maintainers need not worry about.
+
+== Infrastructure Requirements
+
+We will need a number of things to get this going. The low level technical details will be described
```
jglick:
```suggestion
We will need a number of things to get this going. The low-level technical details will be described 
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+can be put in place on `${trusted-ci-server}`, which can provide implicit enforcement of the 
+Jenkins infrastructure team's security best practices. Compliance to these best practices becomes 
+something that plugin maintainers need not worry about.
+
+== Infrastructure Requirements
+
+We will need a number of things to get this going. The low level technical details will be described 
+in a separate Infrastructure Enhancement Request, so, this should be considered a summary for now:
+
+1. The webhook, configured on each participating plugin . Security implications of this are a bit beyond the scope of this document so far.
```
jglick:
```suggestion
1. The webhook, configured on each participating plugin. Security implications of this are a bit beyond the scope of this document so far.
```

jglick: As noted above, I think this is not a requirement, since you could use an organization-wide webhook.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+2. A receiver for the aforementioned webhook, because the `${trusted-ci-server}` will be protected 
+behind a VPN
+3. A trusted Jenkins server for performing builds and deployments. 
+4. `buildPlugin` will need code added to validate `Autorelease.yaml` for correctness, and 
+build the plugin according to the settings described therein
+5. Service account(s), managed by the Jenkins Infrastructure team, which provide secure credentials
```
jglick: FWIW in JEP-305 the deployment is done outside Jenkins, so while the team may manage this service account it is not at risk of compromise from Jenkins itself.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+behind a VPN
+3. A trusted Jenkins server for performing builds and deployments. 
+4. `buildPlugin` will need code added to validate `Autorelease.yaml` for correctness, and 
+build the plugin according to the settings described therein
+5. Service account(s), managed by the Jenkins Infrastructure team, which provide secure credentials 
+to systems such as Artifactory
+
+CAUTION: TODO: Things such as service accounts and new servers will be handled in a separate IEP.
+
+== Testing
```
jglick: I think you are misinterpreting this section of the JEP. This paragraph should be moved to **Reasoning**. The **Testing** section should described how the functionality _of the JEP_ will be tested. In this case I assume it would be interactively, not via automation.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+=== Why YAML?
+YAML is becoming increasingly common in the Jenkins community, for many reasons. YAML is:
+
+* Human readable
+* In use by things like the Kubernetes plugin
+* Easily parsed by any number of publily available libraries
+
+=== Aren't we missing some details about infrastructrue requirements?
+
+In short, yes, we are. This JEP exists to get the conversation started. Once consensus has been reached,
```
jglick: Well, OK, but I find it hard to imagine anyone objecting to the _idea_ of the JEP. Every time the topic is brought up, people ask why we did not have it yesterday. The devil will be in the details of the automation. In particular, the security model needs to be very carefully reviewed.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+This JEP introduces a new system, which enables Jenkins plugin maintainers to add Continuous 
+Delivery to their projects. It is:
```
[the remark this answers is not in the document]
kshultzCB: +1 to that. It's tricky to imagine what this will look like in the end.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+* Opt-in: plugin maintainers can choose to use this
+* Simple to configure: very little needs to be changed in order to enable CD
+* Trusted CI/CD: deployments are produced from a single source of truth, which is managed by
```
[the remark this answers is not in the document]
kshultzCB: Great observation.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+* Opt-in: plugin maintainers can choose to use this
+* Simple to configure: very little needs to be changed in order to enable CD
+* Trusted CI/CD: deployments are produced from a single source of truth, which is managed by 
+the Jenkins project
+
+== Specification
+
+=== Overview
+
+We'll use a short story to illustrate what Autorelease will look like when a plugin maintainer
```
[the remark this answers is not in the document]
kshultzCB: Is it a reasonable term to use? Maybe I should go with something more generic, like;

> We'll use a short story to illustrate what continuous delivery will look like when a plugin maintainer

The config file is called `Autorelease.yaml`. 

I'm probably overthinking this.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+To do so, Karl will:
+
+1. Verify support. Karl's plugin is already in the 
+link:https://github.com/jenkinsci[jenkinsci Organization on GitHub], and is 
+using `buildPlugin` when built on ci.jenkins.io.
+
+2. Choose a branch from which to release. Karl decides he wants to release from the `master` branch, 
+as that is the simplest.
+
+3. Settle on a version numbering scheme. Becuase it's in fairly wide use across the Jenkins
```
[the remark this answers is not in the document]
kshultzCB: Yikes. I've done a lot of that kind of stuff here.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+2. Choose a branch from which to release. Karl decides he wants to release from the `master` branch, 
+as that is the simplest.
+
+3. Settle on a version numbering scheme. Becuase it's in fairly wide use across the Jenkins 
+ecosystem, Karl settles on link:https://semver.org/[semantic versioning].
+
+4. Add a file to the top directory of his repository called `Autorelease.yaml`. 
+
+5. A webhook is configured on Karl's repository on GitHub to point to `{the-webhook-receiver}`
```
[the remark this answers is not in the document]
kshultzCB: 1. The webhook receiver is a computer, which does not exist yet. It'll act as a proxy of sorts, to take these incoming webhooks from commits, and send them to a secure CI server which lives on a VPN. This is in order to keep the CI server in a safe place.

2. My original thought was that the plugin maintainer would configure the webhook on a per-repository basis. That way, if Jane Java decides she wants to use CD on her `jane-plugin`, but release her `old-timey-plugin` from her laptop like she's always done, she can do so.

Maybe an organization-level webhook is the right approach. I hadn't thought of it.  Is there any risk of the webhook receiver being overrun with a huge influx of webhooks, every time someone commits to a repository in the jenkinsci organization? Not necessarily from the standpoint of overwhelming that receiver, but rather, from a logging standpoint. Let's say something goes wrong in this system. The infra team might have an easier time troubleshooting if the only webhooks getting sent are from individual repos, rather than from every repo in the org.

jep/0000/README.adoc
kshultzCB: Fair question. I was using that representation to imply that "trusted-ci-server" was a thing we knew we needed, but didn't have yet. I was probably being overly cutesy there. :) 

If there's a better, more traditional way of implying that, I'm happy to do it that way instead. Perhaps, as you suggest, I could use 

a trusted CI server <sup>2</sup>

And have <sup>2</sup> link to a place in the document where I explain this briefly, and point out that further details will appear in the yet-to-be-written IEP?

jep/0000/README.adoc
kshultzCB: Then I suppose to Artifactory. I didn't know that's how the UC gets populated. Will re-word to reflect that. Maybe 

> After the build is successful, the trusted CI server handles deployment of the built plugin to 
> Artifactory, at which point the existing UC generator makes the plugin available on the update center

If there's documentation on how that UC generator works, I'd (a) love to read that, and (b) be happy to have the words _existing UC generator_ link to that document.

jep/0000/README.adoc
kshultzCB: That's great. Didn't know that was a thing. :1st_place_medal:

jep/0000/README.adoc
kshultzCB: These links are super helpful.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+    </licenses>
+
+    <properties>
+        <revision>3.1.4</revision>
+        <changelist>-SNAPSHOT</changelist>
+        <java.level>8</java.level>
+        <jenkins.version>2.138.4</jenkins.version>
+    </properties>
+
+```
```
kshultzCB: Will need to replace ``` with `----`

jep/0000/README.adoc
kshultzCB: This is related to something that @daniel-beck and I talked about. I understood it in the moment, but the passage of time means I've lost the plot a little. We wanted the number of fields to remain consistent within a given plugin, to avoid having versions sort incorrectly.

jep/0000/README.adoc
kshultzCB: Because the build is getting triggered by a webhook set up specifically for this purpose, the CD builds could be built only on _trusted.ci.jenkins.io_, bypassing the building it twice problem.

jep/0000/README.adoc
kshultzCB: Fair enough, I'll strike it.

jep/0000/README.adoc
kshultzCB: OK, cool. We could adapt that same process here.

jep/0000/README.adoc
kshultzCB: Hopefully, by the same UC generator mechanism (or a similar mechanism) that does this from Artifactory today.

jep/0000/README.adoc
kshultzCB: Yeah, maybe you're right.

jep/0000/README.adoc
kshultzCB: I'll add some words to this effect. A helpful reminder if nothing else.

I mean, hopefully, the maintainers will know what they're getting themselves into, since they had to take the action of adding `Autorelease.yaml` to their repository. But you never know...

jep/0000/README.adoc
kshultzCB: You're right on the money. I misunderstood the intent here. Will move things around.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+[cols="1h,1"]
+|===
+| JEP
+| 0000
+
+| Title
+| Opt-in Continuous Delivery of Jenkins Plugin Releases from Trusted CI
+
+| Sponsor
+| link:https://github.com/daniel-beck[Daniel Beck]
```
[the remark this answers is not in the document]
daniel-beck: I might need to check what I'm signing up for, but sure 😅  I want to see this happen!

jep/0000/README.adoc
bitwiseman: +1 to Jesse's suggestion. 
I think for this section, you can use plain english and use internal linking to sections that describe these in more detail and link out to other docs if needed. (See https://github.com/jenkinsci/jep/blob/master/jep/1/README.adoc#overview for example of what I mean).

jep/0000/README.adoc
bitwiseman: In case some asks why not JSON or Groovy?

jep/0000/README.adoc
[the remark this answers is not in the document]
kshultzCB: I apologize if I misheard you Daniel, but you're stuck with it now. :stuck_out_tongue:

jep/0000/README.adoc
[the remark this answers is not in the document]
kshultzCB: I'm going with 

> We'll use a short story to illustrate what it will look like when a plugin maintainer...

jep/0000/README.adoc
kshultzCB: This might need some tweaking. It's doing "something" but it's not as pretty looking as I was expecting it to be:

image

jep/0000/README.adoc
kshultzCB: Good call. I've stricken it.

jep/0000/README.adoc
daniel-beck: https://github.com/jenkins-infra/backend-update-center2 is the tool for it. We build and run it basically all the time. It downloads the periodically updated Maven index from Artifactory, and uses that to build up a list of plugins and their releases to deliver. There's a bunch of additional features like overrides and performance optimizations here, so it might look pretty weird. That it uses a really obsolete library to do the Maven index stuff doesn't help.

Basically, in terms of deployment, nothing needs to be different, server-side, from today when we get artifacts into Artifactory, the infra knows how to handle it. (Modulo the variable length version numbering scheme we talked about.)

jep/0000/README.adoc
```diff
@@ -0,0 +1,340 @@
...
+ecosystem, Karl settles on link:https://semver.org/[semantic versioning].
+
+4. Add a file to the top directory of his repository called `Autorelease.yaml`. 
+
+5. A webhook is configured on Karl's repository on GitHub to point to a webhook receiver. 
+
+Once ready to create a new release of his plugin, Karl merges some changes to the `master` branch. 
+At which point, the following takes place under the covers: 
+
+1. A webhook is sent to the webhook receier by GitHub
```
jvz:
```suggestion
1. A webhook is sent to the webhook receiver by GitHub
```

jep/0000/README.adoc
daniel-beck: If the pom declares a version prefix that is later amended by the tool, say, `3.1` or `3.1.${build}` or equivalent, we may run into the problem that a "later looking" version number is actually sorting as earlier.

Imagine the fixed version prefix `3.1`, and autorelease adding a build number:

3.1.0
3.1.1
3.1.2
3.1.3

Now, the maintainer changes the apparent semver to `3.1.1`.

3.1.1.0 (or 3.1.1.4, doesn't matter)
3.1.1.1
…

While these releases should be later, they sort earlier. Fixed length version numbers (or "consistent length" per component) addresses this problem.

I also noted this might be circumventable by using a different scheme here – E.g. if something like `3.1.1-b4` or similar would sort differently.

jep/0000/README.adoc
```diff
@@ -0,0 +1,344 @@
...
+1. Verify support. Karl's plugin is already in the 
+link:https://github.com/jenkinsci[jenkinsci Organization on GitHub], and is 
+using `buildPlugin` when built on ci.jenkins.io.
+
+2. Choose a branch from which to release. Karl decides he wants to release from the `master` branch, 
+as that is the simplest.
+
+3. Settle on a version numbering scheme. Karl decides on link:https://semver.org/[semantic versioning].
+
+4. Add a file to the top directory of his repository called `Autorelease.yaml`.
```
jvz: Couldn't this be done as an argument to `buildPlugin()`?

jep/0000/README.adoc
kshultzCB: Good catch. Writing English with a code editor is...well, it makes one appreciate a word processor. :)

jep/0000/README.adoc
daniel-beck: This Jenkins could be on a nonpublic network and report results back using GH statuses. The challenge for incrementals was the deployment of the same artifacts as on public PR build, this doesn't need to have this limitation.

jep/0000/README.adoc
```diff
@@ -0,0 +1,344 @@
...
+1. `pluginName` field must match the repository name
+
+2. `versionNumberFields` must match the `revision` property in `pom.xml`. E.g., if `versionNumberFields` is 
+set to `3` (three fields), and the `revision` in `pom.xml` is set to `3.1.4.5` (four fields), validation 
+will fail.
+
+3. Validation that no extra fields are present in the file. Comments are allowed, but any unexpected 
+lines will cause validation to fail.
+
+If validation of `Autorelease.yaml` fails for any reason, the build is not performed nothing gets deployed,
```
jvz:
```suggestion
If validation of `Autorelease.yaml` fails for any reason, the build is not performed, nothing gets deployed, 
```

jep/0000/README.adoc
daniel-beck: ci.jenkins.io is not involved here at all (except for a CI build of the master branch).

trusted.ci.j.io would be possible, but that instance is a legacy mess.

Would rather go with a shiny new autorelease.ci.j.io on the Jenkins infra VPN.

jep/0000/README.adoc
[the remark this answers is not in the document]
daniel-beck: I think the point was simply that there's no sentence like "The new system described in this JEP will be called Autorelease" or similar.

jep/0000/README.adoc
daniel-beck: I wonder whether an "Autorelease enablement" status check for PRs would be useful here. You see those when you merge.

Alternatively, a possible practice here could be to have a separate 'autorelease' branch – similar to having master/develop, master/autorelease, and releasing from the latter, could be a viable option for people who want to have finer-grained control other releases.

jep/0000/README.adoc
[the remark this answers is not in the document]
jglick: > 1. The webhook receiver is…
> 2. My original thought was…

So write all this in the doc!

> Maybe an organization-level webhook is the right approach…

→ **Reasoning** section

jep/0000/README.adoc
jglick: > nothing needs to be different

That we know of now.

jep/0000/README.adoc
jglick: I have no idea offhand if GitHub’s renderer supports them.

jep/0000/README.adoc
jglick: > Fixed length version numbers addresses this problem.

Until the maintainer

```diff
-versionNumberFields: 2
+versionNumberFields: 3
```

to fix a build error, without considering the consequences on the update center.

> something like `3.1.1-b4` or similar would sort differently

As per [JEP-305](https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc#other-version-schemes), which is not susceptible to this problem because the generated versions always sort _before_ the maintainer-specified string. Of course that particular format is undesirable in this context since these are not “release candidates”, these are full releases—there is no other.

----

My suggestion would be to just get rid of the maintainer-specified component altogether in an autoreleased plugin. It serves no function beyond [pretending to indicate compatibility](https://twitter.com/tyvole/status/1136767636602327042), which is more helpfully and honestly accomplished by writing good release notes. With a minor enhancement to `git-changelist-maven-extension` to allow the `-rc` component to be skipped, the POM can just say

```xml
<version>${changelist}</version>
<properties>
  <changelist>999999-SNAPSHOT</changelist>
</properties>
```

so development builds are all `999999-SNAPSHOT` and incremental builds (including `master` releases) look like `1234.abc123def456`. The more changes you make, the bigger the number, and you never have to pay attention to the version again.

jep/0000/README.adoc
jglick: > except for a CI build of the master branch

That is what I just said. You push to `master` and two Jenkins services initiate builds. Or am I missing something?

jep/0000/README.adoc
jglick: …in which case we do not need another server, or even a new YAML file. ci.jenkins.io will already publish these releases. We would “just” need to patch `update-center2` to also check the Incrementals repository and include the newest release it finds there which

* Matches `\d+[.][a-f0-9]{12}` (i.e., you have opted in by virtue of dropping the manual `revision`).
* Comes from `master` according to [the same logic already used by `incrementals:update`](https://github.com/jenkinsci/incrementals-tools/blob/9a96211ad69fa55599001c9689b5c7e78f9a7f88/lib/src/main/java/io/jenkins/tools/incrementals/lib/UpdateChecker.java#L141-L146).

jep/0000/README.adoc
jglick: IMO people who want finer-grained control over releases can stick with the status quo.

jep/0000/README.adoc
jglick: > patch `update-center2` to also check the Incrementals repository

Better yet, make the `incrementals-publisher` function send artifacts matching these criteria to the regular `releases` repository; or have a separate function or batch process which copies from `incrementals` to `releases`. Thus

* These releases may be made dependencies of other plugins without `-Pconsume-incrementals`.
* We can experiment with GC policies in `incrementals` without fear of dropping bits which were at some point offered on the regular user UC.
* `update-center2` need not be touched—these artifacts would appear in the `releases` index and be published automatically.

jep/0000/README.adoc
daniel-beck: OK, then we agree here.

jep/0000/README.adoc
daniel-beck: > to fix a build error, without considering the consequences on the update center.

From that point on, all autoreleases will necessarily need to fail.

> get rid of the maintainer-specified component altogether in an autoreleased plugin

I expect this will severely limit the acceptance of this system. Not everyone is comfortable with a version number "2412", especially if Autorelease is just being tried out for a given plugin -- no way to get back to version 3.0.x or whatever, after you've released version 2412.

Another problem would be projects like library plugins keeping lockstep with upstream releases, just adding another digit for the Jenkins plugin itself.

daniel-beck: > From that point on, all autoreleases will necessarily need to fail.

… unless accompanied by a major version increase. This is something that a PR build status could tell maintainers.

----

I think we'll have the best chance by imposing as few unnecessary restrictions as on plugin developers, if we want it to gain acceptance. Making the branch and the version scheme configurable is one of those things that should be doable without crazy workarounds and will make it easier for people to give this a try.

jglick: > projects like library plugins keeping lockstep with upstream releases, just adding another digit for the Jenkins plugin itself

```xml
<artifactId>apache-httpcomponents-client-4-api</artifactId>
<version>4.5.5.${changelist}</version>
<properties>
  <changelist>999999-SNAPSHOT</changelist>
</properties>
```

producing versions such as `4.5.5.66.b6af896b0883`, which will sort reasonably. Does mean you would need to pick another flag for `incrementals-publisher`.

daniel-beck: Then the library releases version `4.5.5.1`. What now?

daniel-beck: What is basically what this proposal comes down to, with an extra layer to ensure going from `1.2` to `1.2.1` will not have unintended consequences.

jglick: > you would need to pick another flag for `incrementals-publisher`

For the record: a POM property would not work since the flattened POM omits properties, but you could check for the presence/contents of a named file at that commit.

> What is basically what this proposal comes down to

Really? The proposal as currently written sounds very different to me: a special webhook, a special CI server, etc. The approach I am discussing is something I could probably implement in about a day as a single patch to `incrementals-publisher` (plus prototype usage in some minor plugin I control), given the ability to test changes iteratively like I had from @rtyler when it was initially developed.

> an extra layer to ensure going from `1.2` to `1.2.1` will not have unintended consequences

Is that danger actually specific to this JEP or autoreleasing generally? You can already produce shadowed releases with `maven-release-plugin`, admittedly via a somewhat different kind of source edit. Maybe we should consider other ways of dealing with this class of mistake more broadly, such as having `update-center2` produce a warning (delivered somewhere TBD) when it notices a chronologically new release which is not going to actually be visible in any of the configured update centers. (Not _necessarily_ a mistake, since the maintainer might be producing this release for consumption in some other channel, such as an OEM WAR.)

bitwiseman: @kshultzCB 
This is an example of a discussion that is non-blocking for this first PR.  You'll want to make sure that the salient points are added to the reasoning section, but this is discussion of the design itself not of the structure of the JEP.

bitwiseman:
@jglick 
Thank you for the in-depth review and feedback on this, both from a design and JEP structure perspective.

jep/0000/README.adoc
```diff
@@ -0,0 +1,345 @@
...
+
+1. Verify support. Karl's plugin is already in the 
+link:https://github.com/jenkinsci[jenkinsci Organization on GitHub], and is 
+using `buildPlugin` when built on ci.jenkins.io.
+
+2. Choose a branch from which to release. Karl decides he wants to release from the `master` branch, 
+as that is the simplest.
+
+3. Settle on a version numbering scheme. Becuase it's in fairly wide use across the Jenkins 
+ecosystem, Karl settles on link:https://semver.org/[semantic versioning].
```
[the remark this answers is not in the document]
kshultzCB: Karl is going to use it. :P  (and he also made a minor edit to this line to reflect reality)

jep/0000/README.adoc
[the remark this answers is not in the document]
kshultzCB: Done and done.

jep/0000/README.adoc
kshultzCB: For what it's worth, I've added a link to the UC generator in the appropriate parts of Overview and Detailed Setup. If we find something that needs to be changed, we'll take that as it comes.

jep/0000/README.adoc
kshultzCB: It's still a pretty neat thing, though, thanks for showing that to me.

jep/0000/README.adoc
kshultzCB: This currently reads:

- Once validation passes, a build is performed. If the build passes all its tests, a release is generated. In our example, that release number would be 3.1.4-150-1a2b3c4
- The built plugin gets deployed to Artifactory
- The UC generator makes the plugin available on the Jenkins Update Center, as Karl Plugin version 3.1.4-150-1a2b3c4.

jep/0000/README.adoc
kshultzCB: If I make this change, we lose the ability for readers to find that document:

image

Is that intentional?

jep/0000/README.adoc
kshultzCB: Changed to read as follows:

By using a single system of record for these builds, a service account, maintained by the 
Jenkins CERT team, can be used to access GitHub, deploy to Nexus, and thus deploy to 
the update centers

jep/0000/README.adoc
kshultzCB: I think this is addressed via my additions to the Reasoning section, which now under Reasoning --> Webhooks and the Webhook Receiver.

jep/0000/README.adoc
jglick: Did I not include a URL? I meant to. https://docs.google.com/document/d/1JSxYNI_RuA8ITlxVmxBdFg1A-sOKz-w7a9tzuPfWmr4/edit#heading=h.n2d90ci56ugy

jep/0000/README.adoc
```diff
@@ -0,0 +1,366 @@
...
+Webhooks will be used to trigger the builds on the trusted CI server. The hooks themselves 
+will be configured on a per-repository basis. The reasons for choosing individual hooks, 
+as opposed to an organization wide hook, are:
+
+* Security of the build systems. Because this trusted CI server will reside on a non-public 
+network, there needs to be a proxy server of sorts, which can accept these webhooks and send 
+them, securely, to the trusted CI server.
+
+
+* Traceability. An organization-wide webhook would send those hooks to the receiver far more
```
jglick: As a stylistic matter, please use one source text line per sentence or major phrase. Easier to comment on, diff, etc.

jep/0000/README.adoc
```diff
@@ -0,0 +1,366 @@
...
+* Traceability. An organization-wide webhook would send those hooks to the receiver far more 
+often than is actually necessary. In the event that something goes wrong with this process - 
+the receiver goes down, there is a network outage, etc. - it will be easier for the Infra 
+team to triage problems if they're only looking at hooks from plugins which are intended 
+to be released via CD.
+
+* Flexibility for plugin maintainers. Let's say that Karl maintains two plugins - his archaic 
+`old-timey-plugin`, and his fancy new `karl-plugin` which we've already talked about. If 
+Karl wants to use CD on karl-plugin, but continue to release `old-timey-plugin` from his laptop 
+like he's always done, he can do so, by only setting the webhook up on `karl-plugin`.
```
jglick: Does not make sense to me. You are already proposing to gate autorelease by an explicit flag.

jep/0000/README.adoc
```diff
@@ -0,0 +1,366 @@
...
+* Easily parsed by any number of publicly available libraries
+
+=== Webhooks and the Webhook Receiver
+Webhooks will be used to trigger the builds on the trusted CI server. The hooks themselves 
+will be configured on a per-repository basis. The reasons for choosing individual hooks, 
+as opposed to an organization wide hook, are:
+
+* Security of the build systems. Because this trusted CI server will reside on a non-public 
+network, there needs to be a proxy server of sorts, which can accept these webhooks and send 
+them, securely, to the trusted CI server.
```
jglick: Why does this mean that per-repo webhooks are needed? I am not following the reasoning.

jep/0000/README.adoc
```diff
@@ -0,0 +1,366 @@
...
+
+
+* Traceability. An organization-wide webhook would send those hooks to the receiver far more 
+often than is actually necessary. In the event that something goes wrong with this process - 
+the receiver goes down, there is a network outage, etc. - it will be easier for the Infra 
+team to triage problems if they're only looking at hooks from plugins which are intended 
+to be released via CD.
```
jglick: But conversely you are adding complexity by needing to check whether each repo maintainer remembered to configure this webhook, and configured it correctly. Seems simpler to have an org hook and simply discard events from nonmatching plugins.

jglick:
Oh another thing to think about—interaction with Release Drafter. We would still want to have a **Releases** page for changelogs, even if there is just one bullet point (merged PR) per release. These are based on tags, which would not otherwise exist (unlike with `maven-release-plugin`), so maybe the autorelease process should [create a tag & release](https://developer.github.com/v3/repos/releases/#create-a-release) each time, I guess with the tag name simply being the version number just published (perhaps you need to prefix with `v`). This is something that could easily be done from `incrementals-publisher`, which already has write access so as to set a commit status.

jep/0000/README.adoc
jglick: > you would need to pick another flag for `incrementals-publisher`

Oh! No special flag needed. Only that `-rc` is not present in the version number.

oleg-nenashev:
Is there a dev list thread for it?

kshultzCB:
I have not sent anything to jenkins-dev yet. I'll do once I take off the Draft flag, or maybe before.

jep/0000/README.adoc
```diff
@@ -0,0 +1,366 @@
...
+When Karl merges a commit into the `master` branch, that merge commit is link:https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc#basic-usage[commit number] 150, and has the 
+SHA `1a2b3c4`. The following takes place:
+
+* A webhook is sent from GitHub, to the webhook receiver. That hook is forwarded  to 
+the trusted CI server.
+* `Autorelease.yaml` is validated.
+* Once validation passes, a build is performed. If the build passes all its tests, a release 
+is generated. In our example, that release number would be `3.1.4-150-1a2b3c4`
+* The built plugin gets deployed to Artifactory
+* link:https://github.com/jenkins-infra/update-center2/blob/master/README.md[The UC generator] makes
```
jtnord: I firmly believe NOT creating a tag is a bad idea.

jglick: Can you write up your objections in a form that can be incorporated into **Reasoning**?

jglick:
Picking up https://github.com/jenkinsci/jep/pull/244#discussion_r294029302 as a top-level thread (otherwise it gets lost in the folded section):

My original proposal was actually [INFRA-1570](https://issues.jenkins-ci.org/browse/INFRA-1570), based on GitOps rather than Artifactory index scans, but starting with a scan-driven feature in no way precludes the option of subsequently moving selected plugins into a GitOps-style release: you would merely patch `update-center2` to read their versions from a text file¹ regardless of what the index says.

----

¹Possibly a POM, for ease of Dependabot integration.

jep/0000/README.adoc
```diff
@@ -0,0 +1,366 @@
...
+1. Verify support. Karl's plugin is already in the 
+link:https://github.com/jenkinsci[jenkinsci Organization on GitHub], and is 
+using `buildPlugin` when built on ci.jenkins.io.
+
+2. Choose a branch from which to release. Karl decides he wants to release from the `master` branch, 
+as that is the simplest.
+
+3. Settle on a version numbering scheme. Karl decides on link:https://semver.org/[semantic versioning].
+
+4. Add a file to the top directory of his repository called `Autorelease.yaml`.
```
bitwiseman:
```suggestion
4. Add a file to the top directory of his repository called `Autorelease.yaml`. 

WARNING: *jvz* Couldn't this be done as an argument to buildPlugin()?
```

jep/0000/README.adoc
bitwiseman:
```suggestion
will fail.

[WARNING]
====
*(bitwiseman)* 
There's a long thread here: https://github.com/jenkinsci/jep/pull/244/files#r293534268 that needs to be integrated into this. 





```

jep/0000/README.adoc
bitwiseman:
```suggestion
* Once validation passes, a build is performed. If the build passes all its tests, a release 

WARNING: Some clarification needed based on https://github.com/jenkinsci/jep/pull/244/files#r293536274 .

```

jep/0000/README.adoc
bitwiseman: @jtnord You can provide further feedback after this PR is merged.

jep/0000/README.adoc
kshultzCB: On revisiting this you're right. Was unnecessary and has been deleted.

jep/0000/README.adoc
kshultzCB: Yeah, I need to reword this somehow. The need for a receiver, and the configuration of the hooks themselves (single-repo versus org) aren't the same issue.

jep/0000/README.adoc
kshultzCB: Noted as something to come back to in the latest revision.

jep/0000/README.adoc
kshultzCB: I've moved the `WARNING` section and changed the link so that it points [here](https://github.com/jenkinsci/jep/pull/244/files#r302487016). We'll come back to this.

jglick: The anchor on that URL does not resolve for me, FWIW.

kshultzCB: This is pretty odd. I found what's wrong but I've got no idea how it got that way. URL needs to point to `discussion#anchor` rather than `files#anchor`.

bitwiseman:
Okay, this looks good to go as a Draft JEP-221.

jglick:
Should my counterproposal to the JEP be recorded somewhere?

bitwiseman:
@jglick 
As noted before, the initial PR for a JEP is generally supposed to be a conformance and sanity check, not an extended design discussion.  This JEP definitely ranged into the latter.  I asked @kshultzCB to pull the feedback into annotation sections for later merging.  If that wasn't done, I'm sure it was a simple oversight given the amount of back and forth. 

If you'd like to propose changes to the JEP, please do so.  

If you'd rather create be the sponsor for a JEP that competes with this one, you can do that too.  Your call.

jglick:
According to [JEP-1](https://github.com/jenkinsci/jep/tree/master/jep/1#overview) a draft JEP includes

> a minimal prototype implementation sufficient to convey the viability of the design

and my proposal was for an entirely different design, not a minor refinement of the one proposed in what is now JEP-221, so I guess it would be clearer to file my own—unless major work has already begun on a JEP-221 implementation somewhere that I am not aware of. (There is mention of the need for an IEP, but I do not [see one yet](https://github.com/jenkins-infra/iep/pulls?q=is%3Apr).)

kshultzCB:
[Here's a functioning link to that discussion thread](https://github.com/jenkinsci/jep/pull/244#discussion_r293534268). That's the second time these links have done this in the last hour.

bitwiseman:
@jglick 
Prototype: fair point, my bad. I'll do better next time.  

Let's think about next steps.  

I haven't seen any resistance on the part of @kshultzCB to significant design changes.  Why not propose changes to this JEP for now?

jep/0000/README.adoc
jtnord: developers are used to being able to see and work with tags for releases.  it makes any tooling that needs to checkout a release work without changes (if a plugin is auto released or not should not matter, and it allows you to not worry if something was auto-released or not.  TL;DR this is not the way that maven users expect things to work)

PR #248, opened 2019-10-10
Marking JEP-305 as Final

jglick:
Just noticed this was still in Draft status, which is not right—it is long since implemented and in widespread use. (Certainly there are various open RFEs and the like, but that will always be true.)

### DIFF
--- a/jep/305/README.adoc
+++ b/jep/305/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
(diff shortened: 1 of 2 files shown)


jglick:
Note that there was [discussion of GC](https://groups.google.com/d/msg/jenkinsci-dev/ee3B2ivIgms/kfKXakkGCAAJ) long ago, but we seem to have gotten by without any so far, so I would expect that it would be a new JEP to introduce some form of GC if and when we decide it is warranted. Or we can just assume that storage costs will continue to drop. :-)

PR #249, opened 2019-11-18
Sensible permission segregation for matrix based authorization

mikecirioli:
This proposal, once complete, will describe a more effective implementation of the fine grained permission model for matrix based authorization by introducing a new permission type, `Jenkins.CONFIGURE_JENKINS`, to be used when delegating management of certain aspects of a Jenkins instance that do not present the risk of compromising overall security.


jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.**JEP Template**
+[TIP]
+====
+In this document, all text in a "Tip" block (or inline text with with a ":bulb:" on either side)
+MUST be removed and/or replaced with text appropriate to this JEP before submission.
```
bitwiseman: @mikecirioli  
I assume you will do this before submitting the PR for review?

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+* `Jenkins.RUN_SCRIPTS`
+* `Jenkins.CONFIGURE_UPDATECENTER`
+* `Jenkins.UPLOAD_PLUGINS`
+
+This allows for backwards compatiblilty with plugins while providing motivation to refactor any usages of the deprecated checks in future releases.
+
+Methods that formerly checked for these permissions will instead check for `Jenkins.ADMINISTER`.
+
+All other methods that check for `Jenkins.ADMINISTER` will be evaluated to determine if the use of the new `Jenkins.CONFIGURE` permission type is appropriate.  The general rule of thumb is that configuration item which does not let you change the underlying security model (either directly, or through a malicious plugin or groovy script), or change fundamental configurations such as hostnames, proxy configuration, etc. are likely candidates to use the `Jenkins.CONFIGURE`.
```
daniel-beck: Sentence-per-line formatting would make it much easier to comment on specific statements.

daniel-beck:
I get the impression that this JEP combines two essentially unrelated changes:

* Deprecation of "dangerous" permissions since they aren't used in practice anyway.
* Introduction of a new "Administer Light" permission called "Configure Jenkins"

Why are these seemingly unrelated changes combined into a single JEP? The document does not explain this.

jep/0000/README.adoc
mikecirioli: yes, there is still a bit of work to be done before i take this PR out of draft state.

jglick:
> Why are these seemingly unrelated changes combined into a single JEP?

They feel related to me.

daniel-beck:
> They feel related to me.

Perhaps _unrelated_ is too strong, but they do not appear to be dependent on each other, and explanations, motivations etc. seem to be separate. I cannot think of a reason that would prevent only doing one and not the other.

Hence the combination in one JEP seems unnecessary, complicating providing feedback on specific parts of the proposal. Doing both makes sense IMO, but they seem independent.

mikecirioli:
> > They feel related to me.
> 
> Perhaps _unrelated_ is too strong, but they do not appear to be dependent on each other, and explanations, motivations etc. seem to be separate. I cannot think of a reason that would prevent only doing one and not the other.
> 
@daniel-beck / @jglick  - would it be better to just go ahead and create a PR which:

-  flags those specific permissions (`RUN_SCRIPTS`, `CONFIGURE_UPDATECENTER`, and `UPLOAD_PLUGINS`) as deprecated
- update any permissions checks for `RUN_SCRIPTS`, `CONFIGURE_UPDATECENTER`, and `UPLOAD_PLUGINS` to instead check for `Jenkins.ADMINISTER`
- and remove any references in this jep?

see https://github.com/jenkinsci/jenkins/pull/4365 for an example draft PR

daniel-beck:
Maybe? Seems more beneficial to have two separate JEPs from this. AFAIUI, it would basically be this PR, split in two. We still have security warnings up for plugins that mistake Overall/Administer for Overall/RunScripts from old advisories, so ideally the deprecation would be better reasoned and documented than one line in the changelog.

mikecirioli:
@daniel-beck - do you feel strongly that deprecating the permissions requires a JEP?  They are effectively treated as `ADMINISTER` today?

oleg-nenashev:
> do you feel strongly that deprecating the permissions requires a JEP? They are effectively treated as ADMINISTER today?

I do not think JEP is needed here, we already had security advisories and disabled them by default in top authorization plugins.

But it worth having a separate dev list discussion to see whether there is a consensus there

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+When using matrix-based access control the permission types `Jenkins.RUN_SCRIPTS`, `Jenkins.CONFIGURE_UPDATECENTER`, and `Jenkins.UPLOAD_PLUGINS` provide a false sense of security, as a user is able to escalate this limited privilege to `Jenkins.ADMINISTER`.  These permissions are being marked as `@Deprecated` and are now consolidated as `Jenkins.ADMINISTER`.  
+
+A new permission type, `Jenkins.CONFIGURE` (implied by `Jenkins.ADMINISTER`)has been created for actions that allow for the management of Jenkins configuration items that do not require `Jenkins.ADMINISTER`.  This new permission is used to create a clear separation between settings that impact the entirety of a Jenkins instance (particularly with regards to security) and those that can be used to manage other aspects of an installation.
```
jglick:
```suggestion
A new permission type, `Jenkins.CONFIGURE_JENKINS` (implied by `Jenkins.ADMINISTER`) has been created for actions that allow for the management of Jenkins configuration items that do not require `Jenkins.ADMINISTER`.  This new permission is used to create a clear separation between settings that impact the entirety of a Jenkins instance (particularly with regards to security) and those that can be used to manage other aspects of an installation. 
```

according to the prototype implementation, though I prefer `Jenkins.CONFIGURE` as stated here:

* `CONFIGURE` is consistent with permissions of the same name in `Item`, `Computer`, etc.
* `Jenkins.CONFIGURE_JENKINS` is redundant.

jep/0000/README.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+== Abstract
+
+When using matrix-based access control the permission types `Jenkins.RUN_SCRIPTS`, `Jenkins.CONFIGURE_UPDATECENTER`, and `Jenkins.UPLOAD_PLUGINS` provide a false sense of security, as a user is able to escalate this limited privilege to `Jenkins.ADMINISTER`.  These permissions are being marked as `@Deprecated` and are now consolidated as `Jenkins.ADMINISTER`.  
+
+A new permission type, `Jenkins.CONFIGURE` (implied by `Jenkins.ADMINISTER`)has been created for actions that allow for the management of Jenkins configuration items that do not require `Jenkins.ADMINISTER`.  This new permission is used to create a clear separation between settings that impact the entirety of a Jenkins instance (particularly with regards to security) and those that can be used to manage other aspects of an installation. 
+
+== Specification
+
+The following permission types are now marked as `@Deprecated` and `@Restricted(NoExternalUse.class)`:
```
jglick: Remove `@Restricted(NoExternalUse.class)` please. Just `@Deprecated`.

jep/0000/README.adoc
jglick: Yes please! See https://github.com/jenkinsci/jep/tree/master/jep/1#asciidoc-style-and-linter and specifically https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

jglick:
> I cannot think of a reason that would prevent only doing one and not the other.

There is no technical obstacle, but conceptually the motivation to me for deprecating these old permissions _now_ is that the misconception behind their creation has led to the introduction of the new permission which is supposed to replace them. I find it easier to review and comment on a single, coherent change to how Jenkins global permissions works. That said, I am not opposed to doing the deprecation as a preliminary step and removing it from this JEP if other reviewers find that more comfortable.

jep/0000/README.adoc
mikecirioli: 👍

jep/0000/README.adoc
mikecirioli: I have made the decision to remove all of the `@Deprecated` related text and keep this JEP focused on the addition of the new permission.  We will propose changes related to deprecating `RUN_SCRIPTS`, etc outside of this JEP

jep/0000/README.adoc
```diff
@@ -0,0 +1,325 @@
...
+Existing tests that validate permissions, or make assumptions about the permissions being used, will be updated to conform to the new permission model.
+Additional tests will be written that validate the new permission type cannot be used when the more restrictive `Jenkins.ADMINISTER` is needed.
+
+== Prototype Implementation
+[NOTE]
+The prototype code found below will be moved to a proper _DRAFT_ pull request in the near future
+
+A WIP prototype implementation can be found at:
+
+* link:https://github.com/mikecirioli/jenkins/tree/FGP[WIP branch with proposed changes]
```
jglick:
```suggestion
* link:https://github.com/jenkinsci/jenkins/compare/master...mikecirioli:FGP[WIP branch with proposed changes]
```

(though a draft PR would be even better)

mikecirioli: We will create a `Draft` PR in the next couple of days

aHenryJard: the draft PR is here https://github.com/jenkinsci/jenkins/pull/4374

bitwiseman:
@daniel-beck 
So, do you feel strongly that this should be two separate JEPs? I know you've said it twice, but people have offered more information since. You are the reasonable BDFL-Delegate for this JEP, so it is your call. You just need to say it one more time. 😄

mikecirioli:
As part of this proposal we have been struggling a bit to find the right "name" to describe this new permission type.  Currently, we are thinking about creating a new `Permission Group` called `Restricted Administer` in order to provide some contextual meaning to the permissions it contains.  Initially there will be only the proposed permission, `Configure`, but there are a number of other potentially similar permission types that would fit in this category as well (although not in scope for this JEP).  Below is a screenshot to help illustrate what we are considering.
  
permission_idea

oleg-nenashev:
I plan to merge it next week as a draft if no negative feedback

oleg-nenashev:
I kindly suggest that one of security team members is assigned as a BDFL Delegate. CC @daniel-beck who could suggest the best candidate

jglick:
> if no negative feedback

If you are an editor, you do not need to wait for feedback—so long as the JEP is well-formed, it can be filed as a draft. This does not imply acceptance of the _proposal_.

oleg-nenashev:
It is just a template GitHub message @jglick :)

mikecirioli:
@oleg-nenashev I have an open request with the security team to review and (preferably) assume the role of BDFL Delegate

PR #252, opened 2019-12-12
README: Remove the reference to defunct JEP meetings

oleg-nenashev:
We do not longer have JEP Meetings, and there is no immediate plan to remove them. I suggest just removing the sentence for now


oleg-nenashev:
ping @bitwiseman @jenkinsci/jep-editors

oleg-nenashev:
Thnks!

PR #253, opened 2019-12-12
Create CODEOWNERS

oleg-nenashev:
Just because we can


oleg-nenashev:
I will merge it so that the process proceeds for new JEP PRs. It does not essentially change anything except explicitly requesting reviews from the editorial team

PR #256, opened 2020-01-09
Update JEP-223 to better reflect the intent of the new permission 

mikecirioli:


oleg-nenashev:
Sorry for the wrong ID in the body

PR #260, opened 2020-01-15
Address JEP-223 feedback

mikecirioli:
This PR addresses the following feedback from @daniel-beck and @jvz :

- This is a fairly complex change with no commitment to address regressions (like JEP-200 did). 
- If there are still plugins using the deprecated Plugin#configure  mechanism, their configuration appears to be always hidden from Jenkins/Configure via GlobalPluginConfiguration#configure . We're probably good with that?

> Plugins that contribute to the settings on on the Configure Jenkins page should carefully consider if allowing a user with only Jenkins.CONFIGURE could result in an unintended privelege escalation.

- This sounds like a fairly onerous requirement for plugins. Based on the numerous published security advisories just around plugins not bothering with a permission check in the first place, this sounds like a fairly large hole. How could we mitigate that?

### DIFF
--- a/jep/223/README.adoc
+++ b/jep/223/README.adoc
@@ -222,9 +222,9 @@ Since the granting of the `Jenkins.CONFIGURE` is optional, administrators who fi
 While we do not anticipate that this change will introduce any regressions, we do acknowledge that some cases may be found after the initial release that may result in confusing or broken behavior for some plugins (ex. a configure link is shown for some plugin, but clicking it results in a permission failure when actually accessing its management link).  
 Because of this, we (the JEP-223 sponsors) are committed to the following support policy:
 
-* For issues raised against Jenkins Core, we will review and fix (or provide justification for not fixing) problems related to this change
-* For issues raised against plugins that are bundled with Jenkins, we will review and fix (or provide justification for not fixing) problems related to this change
-* For issues raised against plugins that are not bundled with Jenkins, we will review and determine the level of impact caused by the regression.  If the issued is determined to create a significant regression we will commit to providing a fix or workaround.
+* For issues raised against Jenkins Core, we will review and fix (or provide justification for not fixing) problems related to this change within 3 months of the LTS release
+* For issues raised against plugins that are installed by default through the setup wizard, we will review and fix (or provide justification for not fixing) problems related to this change within 3 months of the LTS release
+* For issues raised against plugins that are not installed by default through the setup wizard, we will review and determine the level of impact caused by the regression.  If the issued is determined to create a significant regression we will commit to providing a fix or workaround.
 
 [NOTE]
 ====
--- a/jep/223/README.adoc
+++ b/jep/223/README.adoc
@@ -220,10 +220,16 @@ No existing permission types are being removed at this point, which will allow f
 Removing code references to the "dangerous" permissions is outside the scope of this JEP.
 Since the granting of the `Jenkins.CONFIGURE` is optional, administrators who find the new permission does not provide value for them can simply continue to grant `Jenkins.ADMINISTER` to users who need overall administrator access (`Jenkins.CONFIGURE` is implied by `Jenkins.ADMINISTER`).
 While we do not anticipate that this change will introduce any regressions, we do acknowledge that some cases may be found after the initial release that may result in confusing or broken behavior for some plugins (ex. a configure link is shown for some plugin, but clicking it results in a permission failure when actually accessing its management link).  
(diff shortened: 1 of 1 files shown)


mikecirioli:
@aHenryJard / @EstherAF  - please review

jep/223/README.adoc
```diff
@@ -218,6 +218,12 @@ For the most current discussions on these points (and more) please refer to link
...
+Since the granting of the `Jenkins.CONFIGURE` is optional, administrators who find the new permission does not provide value for them can simply continue to grant `Jenkins.ADMINISTER` to users who need overall administrator access (`Jenkins.CONFIGURE` is implied by `Jenkins.ADMINISTER`).
+While we do not anticipate that this change will introduce any regressions, we do acknowledge that some cases may be found after the initial release that may result in confusing or broken behavior for some plugins (ex. a configure link is shown for some plugin, but clicking it results in a permission failure when actually accessing its management link).  
+Because of this, we are committed to the following support policy:
+* For issues raised against Jenkins Core, we will review and fix (or provide justification for not fixing) problems related to this change
+* For issues raised against plugins that are bundled with Jenkins, we will review and fix (or provide justification for not fixing) problems related to this change
+* For issues raised against plugins that are not bundled with Jenkins, it will be up to the maintainers of those plugins to address any found issues
```
oleg-nenashev: Would need some clarification. If a serious regression is caused by the change, my expectation would be that the JEP sponsors are on the hook to fix it

jep/223/README.adoc
```diff
@@ -218,6 +218,18 @@ For the most current discussions on these points (and more) please refer to link
...
 Removing code references to the "dangerous" permissions is outside the scope of this JEP.
+Since the granting of the `Jenkins.CONFIGURE` is optional, administrators who find the new permission does not provide value for them can simply continue to grant `Jenkins.ADMINISTER` to users who need overall administrator access (`Jenkins.CONFIGURE` is implied by `Jenkins.ADMINISTER`).
+While we do not anticipate that this change will introduce any regressions, we do acknowledge that some cases may be found after the initial release that may result in confusing or broken behavior for some plugins (ex. a configure link is shown for some plugin, but clicking it results in a permission failure when actually accessing its management link).  
+Because of this, we (the JEP-223 sponsors) are committed to the following support policy:
+
+* For issues raised against Jenkins Core, we will review and fix (or provide justification for not fixing) problems related to this change
+* For issues raised against plugins that are bundled with Jenkins, we will review and fix (or provide justification for not fixing) problems related to this change
```
daniel-beck: There is no such thing in any sense that would be useful to anyone. Perhaps this?

```suggestion
* For issues raised against plugins that are installed by default through the setup wizard, we will review and fix (or provide justification for not fixing) problems related to this change
```

jep/223/README.adoc
```diff
@@ -218,6 +218,18 @@ For the most current discussions on these points (and more) please refer to link
...
+* For issues raised against plugins that are not bundled with Jenkins, we will review and determine the level of impact caused by the regression.  If the issued is determined to create a significant regression we will commit to providing a fix or workaround.
+
+[NOTE]
+====
+TBD: The definition of a "serious regression"
```
daniel-beck: Also time limit. 1 year should be plenty.

jep/223/README.adoc
```diff
@@ -218,6 +218,18 @@ For the most current discussions on these points (and more) please refer to link
 
 No existing permission types are being removed at this point, which will allow for existing plugins that make use of them to continue to function.
 Removing code references to the "dangerous" permissions is outside the scope of this JEP.
+Since the granting of the `Jenkins.CONFIGURE` is optional, administrators who find the new permission does not provide value for them can simply continue to grant `Jenkins.ADMINISTER` to users who need overall administrator access (`Jenkins.CONFIGURE` is implied by `Jenkins.ADMINISTER`).
+While we do not anticipate that this change will introduce any regressions, we do acknowledge that some cases may be found after the initial release that may result in confusing or broken behavior for some plugins (ex. a configure link is shown for some plugin, but clicking it results in a permission failure when actually accessing its management link).
```
daniel-beck: Unlikely that this will really happen as a regression (as the link should have been governed by the same permission as the resulting configuration page).

jep/223/README.adoc
oleg-nenashev: Maybe even an overkill. For JEP-200 the commitment was 3 months after the LTS release with a fix

PR #261, opened 2020-01-21
System read permission JEP

timja:


jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+
+|===
+
+== Abstract
+
+This JEP introduces a method to allow system 'configuration' to be viewed but not changed.
+
+== Specification
+
+Any methods that check for `Jenkins.ADMINISTER`, `Jenkins.RUN_SCRIPTS`, `PluginManager.CONFIGURE_UPDATECENTER`, and `PluginManager.UPLOAD_PLUGINS` will be evaluated to determine if the use of the new `Jenkins.SYSTEM_READ` permission type is appropriate.
```
mikecirioli: @timja As i understand it, this proposal potentially conflicts with the one proposed by [JEP-223](https://github.com/jenkinsci/jep/tree/master/jep/223).  I _think_ that when a user goes to save any management page, we would end up with the desired result (ie. a user with `Jenkins.CONFIGURE` would be able to persist some parts of the config they were shown, and a user with `Jenkins.SYSTEM_READ` would not be able to persist anything), but this user experience may be confusing to most users.

timja: I think we can make it work, possibly some or conditions and additional permission checks may be required in the UI.

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+
+== Specification
+
+Any methods that check for `Jenkins.ADMINISTER`, `Jenkins.RUN_SCRIPTS`, `PluginManager.CONFIGURE_UPDATECENTER`, and `PluginManager.UPLOAD_PLUGINS` will be evaluated to determine if the use of the new `Jenkins.SYSTEM_READ` permission type is appropriate. 
+
+The general rule of thumb is that all views should be readable by `Jenkins.SYSTEM_READ`, unless they give no benefit to the user.
+An example of a non beneficial view is the 'Script console' view, this is too dangerous for non administrators,
+and there is no useful information on here so the permissions on this view won't be changed.
+
+By default, all of the configuration sections in the `/manage` page require `Jenkins.ADMINISTER`, this will be changed to `Jenkins.SYSTEM_READ`. This is safe as plugins will have additional permission checks on `Jenkins.ADMINISTER` on form submission, to prevent existing CSRF attacks.
```
mikecirioli: What would the user experience look like in this scenario?  Do they see an error page when trying to save their changes, or does it just silently not persist their changes?

timja: For pages that have been updated they won't see the apply button.
If the page hasn't been updated they will get a permission denied error when they click the apply button

jep/0000/README.adoc
EstherAF: @timja @mikecirioli 
Following this matter, because the two permissions are not directly related in the permission hierarchy, the only solution I can think is implementing a way of checking if the user has at least one of them. But I feel like we are hitting a design limitation here, and this solution might be a "dirty" workaround.

Thoughts? Maybe there is a simpler solution :D

Apologies if this topic has been already covered in another place

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+[cols="1h,1"]
+|===
+| JEP
+| 0000
+
+| Title
+| Readonly system configuration
+
+| Sponsor
+| link:https://github.com/timja[timja]
```
oleg-nenashev: Full names are recommended

jep/0000/README.adoc
```diff
@@ -0,0 +1,227 @@
...
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+| JIRA
+| https://issues.jenkins-ci.org/browse/JENKINS-12548[JENKINS-12548]
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
```
oleg-nenashev: Would be great to kick-off the dev list discussion

jep/0000/README.adoc
timja: I'll have to look up the code later on but I think I have seen some places that allow multiple permission with an or condition

jep/0000/README.adoc
mikecirioli: Attempting to kickstart this discussion on the dev mailing lists --->  https://groups.google.com/d/msg/jenkinsci-dev/SHWZiMfbPEs/ZZEj00OsEAAJ

PR #269, opened 2020-02-13
JEP-223 Add detail on how ManagementLink changes with plugin usage analyse.

aHenryJard:


jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
+
+|script-security-plugin
+|No change
+|getRequiredPermission() is overrided with `RUN_SCRIPT` in the ManagementLink
+
+|active-choices-plugin
+|No change
+|Based on scriptler-plugin
```
Wadeck: Scriptler is an optional dependency but not called there.
The requiredPermission is asked on ScriptlerScript.DescriptorImpl but this class and its hierarchy are missing the desired method/field.
=> That never worked since the initial commit and it's just a bug in their plugin.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+
+|global-build-stats-plugin
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
+
+|selenium-plugin
+|No change
+|getRequiredPermission() is overrided with a custom permission in the ManagementLink
+|===
+
```
Wadeck: When looking for "equiredPermission" in the ecosystem, I got 191 occurences in 131 different files (java, groovy, jelly. It seems you did not cover all of them.

Second item on my list => build-metrics, two bugs, they expected to use the requiredPermission from ManagementLink in their views but the "it" was not the correct one and thus no impact.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No change
+|getRequiredPermission() is overrided with `RUN_SCRIPT` in the ManagementLink
+
+|active-choices-plugin
+|No change
+|Based on scriptler-plugin
+
+|docker-plugin
+|Change
+|Does not override getRequiredPermission() in the ManagementLink and use it in jelly. The management page will be restricted to `ADMINISTER`.
```
Wadeck: Due to [DockerManagement#getTarget](https://github.com/jenkinsci/docker-plugin/blob/master/src/main/java/com/nirima/jenkins/plugins/docker/DockerManagement.java#L82-L85), requiring admin is the expected behavior.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
+
+|ssh-slaves-plugin
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
+
+|docker-traceability-plugin
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
```
Wadeck: FTR it's a RootAction. The difference is that the RootAction provides a global URL reachable without any item.

jep/223/README.adoc
Wadeck: - dynamicparameter, two occurences:
The requiredPermission is asked on ScriptlerChoiceParameterDefinition.DescriptorImpl and ScriptlerStringParameterDefinition.DescriptorImpl but the classes and their respective hierarchies are missing the desired method/field.
=> That never worked since the initial commit and it's just a bug in their plugin.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No change
+|Based on scriptler-plugin
+
+|periodic-jobs-balancer-plugin
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
+
+|gerrit-trigger-plugin
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
```
Wadeck: FTR, mix of ManagementLink, Action and RootAction in the usage of requiredPermission.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
+
+|rebuild-plugin
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
+
+|global-build-stats-plugin
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
```
Wadeck: The GlobalBuildStatsPlugin is a Plugin, it has a method get `requiredPermission`, but its internal ManagementLink has nothing there (and is not used in the view)

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No change
+|Based on scriptler-plugin
+
+|docker-plugin
+|Change
+|Does not override getRequiredPermission() in the ManagementLink and use it in jelly. The management page will be restricted to `ADMINISTER`.
+
+|libvirt-slave-plugin
+|Change
+|Does not override getRequiredPermission() in the ManagementLink and use it in jelly. The management page will be restricted to `ADMINISTER`.
```
Wadeck: Due to [VirtualMachineManagement#getTarget](https://github.com/jenkinsci/libvirt-slave-plugin/blob/master/src/main/java/hudson/plugins/libvirt/VirtualMachineManagement.java#L69-L72), it's expected to be restricted to admin.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+
+|build-with-parameters-plugin
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
+
+|metadata-plugin
+|No related
+|The requiredPermission is used from an Action class, not a ManagementLink class.
```
Wadeck: FTR both are RootAction,

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
+
+|scriptler-plugin
+|No change
+|Manages its own permission in the ManagementLink
+
+|p4-plugin
+|No change
+|Based on scriptler-plugin
```
Wadeck: Nothing related to scriptler, ReviewAction is an Action and P4Hook an UnprotectedRootAction, so this JEP has no impact here.

jep/223/README.adoc
```diff
@@ -269,6 +269,108 @@ The intent of this proposal is to improve overall security for Jenkins instances
...
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
+
+|global-build-stats-plugin
+|No change
+|getRequiredPermission() is overrided with `ADMINISTER` in the ManagementLink
+
+|selenium-plugin
+|No change
+|getRequiredPermission() is overrided with a custom permission in the ManagementLink
```
Wadeck: FTR PluginImpl is not a ManagementLink, ServiceManagementAction an Action and SeleniumGlobalConfiguration a Describable.

jep/223/README.adoc
Wadeck: - support-core, lots of getRequiredPermissions, due to their Component approach, nothing impacting.

Wadeck: - workflow-cps, same as support as it provides components to that plugin

aHenryJard: Thank you for your feedback @Wadeck I am going to add this last 4 plugins.

jep/223/README.adoc
aHenryJard: Yes sorry I did a shortcut and put the same text for both Action and RootAction classes.

jep/223/README.adoc
aHenryJard: When is the getTarget used in the lifecycle of the management page ? Before displaying the page or on save ?

Wadeck: If you see a class in Jenkins ecosystem, implementing StaplerProxy with its "getTarget" method, this is used during Stapler routing process. When the routing arrives on the class (as a fragment/step inside the "URL path" traversal), it will first check the getTarget() and use the return of this method. This could be used to check permission (throwing AccessDeniedException) or changing the target for this fragment.

aHenryJard:
@oleg-nenashev I addressed nearly all comment and @Wadeck has reviewed the changes. I think that you can ship it as it is now. I my mind it's better to have a 90% coverage analyse than zero since a lot of people are asking for clarifications on that ManagementLink.

PR #280, opened 2020-04-21
Mark JEP-202 as Final

jglick:
Long since completed.

### DIFF
--- a/jep/202/README.adoc
+++ b/jep/202/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
-| Accepted :ok_hand:
+| Final :lock:
 
 | Type
 | Standards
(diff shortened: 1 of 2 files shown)


jglick:
`list` and `toExternalURL` could be made final. `mode` and `readLink`, less sure. Was on my to-do list in the longer term, did not realize that had much to do with JEP status.

oleg-nenashev:
This is definitely a gray zone in the JEP process, because there was no Beta API support when it was introduced. I would vote for the following approach:

* A proposal might be `accepted` when there are Beta APIs
* For a `final` proposal Beta APIs should be either promoted to final ones or removed. It might be also possible to exclude these APIs to a separate experimental feature if it is reflected in the JEP (no need to create JEP for that)

jglick:
OK, would need to spend some time looking into the possibilities here.

jglick:
See https://github.com/jenkinsci/jenkins/pull/4974.

oleg-nenashev (approved):
Fine with me. We have 2+ reference implementations, including one created by external developers (Azure artifact storage)

jglick:
> 2+ reference implementations

Well. There is one “reference implementation” of the SPI—`artifact-manager-s3`—and some reference implementations of the API—`workflow-basic-steps` (IIRC) and `copyartifact`. I have not personally reviewed the Azure implementation to see how much of the SPI is used and how, but I am unaware of any complaints at least. CloudBees CI also includes additional uses of the API.

timja (approved):

PR #281, opened 2020-04-21
Mark JEP-222 as Final

jglick:
In an LTS; no further work planned for now.

### DIFF
--- a/jep/222/README.adoc
+++ b/jep/222/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
-| Draft :speech_balloon:
+| Final :lock:
 
 | Type
 | Standards
--- a/jep/README.adoc
+++ b/jep/README.adoc
@@ -187,7 +187,7 @@
 | link:https://github.com/daniel-beck[Daniel{nbsp}Beck]
 | TBD
 
-| Draft{nbsp}:speech_balloon:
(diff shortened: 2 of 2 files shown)


oleg-nenashev (dismissed):
Needs APIs to be promoted to the Final state

oleg-nenashev:
Also, we cannot easily move to Final without having a BDFL delegate IMHO. Let's take it to the governance meeting

jglick:
> Needs APIs to be promoted to the Final state

Really? The APIs are there in case a legitimate usage from a plugin is found. To date there is none, though there is a _potential_ usage in a proprietary plugin I know of.

jglick:
> without having a BDFL delegate

Oh is that still a thing when we have no active BDFL? In that case maybe never mind, I do not _really_ care what the JEP status is.

jglick:
> there is a _potential_ usage in a proprietary plugin I know of

@Vlatombe has implemented that successfully, for the record.

jglick:
What about Accepted state?

timja (approved):
JEP 1 states:

>  **<<final, Final Status>>** - When the implementation is complete
  and all changes have been published
  or otherwise incorporated into the appropriate code repositories,
  the JEP status is changed to "<<final, Final>>"
  (or in some cases "<<active, Active>>").
  The JEP is done.

This meets that criteria, will hold off a bit for any more feedback from Oleg before merging, (ping me here if there's no response and I forget)

PR #282, opened 2020-04-21
Mark Blue Ocean JEPs as deferred (JEP-203, JEP-204) 

oleg-nenashev:
Based on https://groups.google.com/forum/#!msg/jenkinsci-dev/zO9VFNxhF-o/EpjdA8-jCgAJ , I do not think this JEPs are going to proceed in foreseeable future. I suggest to explicitly defer them. CC @bitwiseman @jphartley @imeredith who could comment on that.

I am not sure about JEP-215, needs additional feedback

### DIFF
--- a/jep/203/README.adoc
+++ b/jep/203/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
-| Draft :speech_balloon:
+| Deferred :hourglass:
 
 | Type
 | Standards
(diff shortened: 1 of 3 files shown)


jphartley:
At least from CloudBees perspective our policy around Blue Ocean is that we have moved it into maintenance mode while we focus on developing a new UI which will incorporate the best parts of Blue Ocean (mainly pipeline visualization). This means that we will resolve major defects, security issues and worthy community PRs. We don't however support major new development initiatives.

PR #285, opened 2020-05-18
[JEP-223] Using Epic JIRA instead of the first PR Jira

aHenryJard:
To be more complete and accurate about the JEP, it seems better to use the EPIC link as Jira link.


oleg-nenashev:
thanks!

PR #291, opened 2020-06-17
[JEP-226] Improve section on fingerprint cleanup

stellargo:
Why am I doing this? I now realise that it does not make sense to keep the build pointers to a fingerprint updated if we have to run a periodic job anyway. Might as well use it to update the build pointers, just like the current implementation does it.

I have also proposed revamping the current fingerprint cleanup to use the new API, so it becomes homogenous.

@oleg-nenashev @afalko @mikecirioli  Would love to hear your thoughts, as this is highly important for the next steps :)

### DIFF
--- a/jep/226/README.adoc
+++ b/jep/226/README.adoc
@@ -122,12 +122,22 @@ For an external storage implementation, this creates a problem because the exter
 
 This proposal offers the following solution:
 
-* When a build/job is deleted from Jenkins, a subsequent update is performed to the fingerprint to reflect this change
 * An async periodic work is run to check if any fingerprint has no jobs/builds associated with it and delete it. (But, it would ensure that if there is a deletion blocking `FingerprintFacet`, then the fingerprint would not be deleted.)
 
-The above solution is provided to the user as an option, and can be toggled because it may be the case that storing data (i.e. fingerprints) might be cheaper than to do consistent updates in the database.
+New methods to be introduced inside the `FingerprintStorage` class:
 
-We do not propose to revamp the `FingerprintCleanupThread` to use the above logic at the moment.
+* `void setFingerprintCleanupThread(boolean toggle, @CheckForNull how_often_to_run_it)`
+
+** FingerprintCleanupThread can be enabled or disabled. External storages can use this method. We can set the trigger interval also here.
+
+* `List<String> getAllFingerprintIds()`
+
+** Returns all the fingerprint ids currently stored by the particular Jenkins instance.
+
+We propose to revamp the `FingerprintCleanupThread` to use the new API.
+The new `FingerprintCleanupThread` will use `FingerprintStorage#getAllFingerprintIds()` to obtain the list of fingerprints to check. If all the associated builds to a fingerprint have been deleted, `FingerprintStorage#delete(String id)` will be called in order to delete the fingerprint.
+
+The above solution is provided to the user as an option, and can be toggled because it may be the case that storing data (i.e. fingerprints) might be cheaper than to do consistent updates in the database.
 
 ==== External to Local Migration (out of scope for MVP)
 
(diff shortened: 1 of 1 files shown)


jep/226/README.adoc
```diff
@@ -122,12 +122,22 @@ For an external storage implementation, this creates a problem because the exter
 
 This proposal offers the following solution:
 
-* When a build/job is deleted from Jenkins, a subsequent update is performed to the fingerprint to reflect this change
 * An async periodic work is run to check if any fingerprint has no jobs/builds associated with it and delete it. (But, it would ensure that if there is a deletion blocking `FingerprintFacet`, then the fingerprint would not be deleted.)
 
-The above solution is provided to the user as an option, and can be toggled because it may be the case that storing data (i.e. fingerprints) might be cheaper than to do consistent updates in the database.
+New methods introduced inside the `FingerprintStorage` class:
+
+* `void setFingerprintCleanupThread(boolean toggle, @CheckForNull how_often_to_run_it)`
```
afalko, line 129: Document that toggle == true means "on" and false means "off"

afalko: What object is "how_often_to_run" ? Should this be a cron?

stellargo:
@afalko @oleg-nenashev @mikecirioli After discussions we had in our meeting, I have revamped the design. Your reviews would be greatly helpful. TL;DR Fingerprint cleanup thread has a default builtin implementation which can be overridden by plugins to provide more efficient implementations (say using cursors). The whole thing can be toggled on or off.

jep/226/README.adoc
```diff
@@ -122,12 +122,25 @@ For an external storage implementation, this creates a problem because the exter
 
 This proposal offers the following solution:
 
-* When a build/job is deleted from Jenkins, a subsequent update is performed to the fingerprint to reflect this change
 * An async periodic work is run to check if any fingerprint has no jobs/builds associated with it and delete it. (But, it would ensure that if there is a deletion blocking `FingerprintFacet`, then the fingerprint would not be deleted.)
 
-The above solution is provided to the user as an option, and can be toggled because it may be the case that storing data (i.e. fingerprints) might be cheaper than to do consistent updates in the database.
+New methods introduced inside the `FingerprintStorage` class:
+
+* `void setFingerprintCleanupThread(boolean toggle, @CheckForNull int recurrence_period)`
```
oleg-nenashev, line 129: May not be a thread actually

```suggestion
* `void setupPeriodicFingerprintCleanup(boolean toggle, @CheckForNull int recurrence_period)`
```

jep/226/README.adoc
```diff
@@ -122,12 +122,25 @@ For an external storage implementation, this creates a problem because the exter
 
 This proposal offers the following solution:
 
-* When a build/job is deleted from Jenkins, a subsequent update is performed to the fingerprint to reflect this change
 * An async periodic work is run to check if any fingerprint has no jobs/builds associated with it and delete it. (But, it would ensure that if there is a deletion blocking `FingerprintFacet`, then the fingerprint would not be deleted.)
 
-The above solution is provided to the user as an option, and can be toggled because it may be the case that storing data (i.e. fingerprints) might be cheaper than to do consistent updates in the database.
+New methods introduced inside the `FingerprintStorage` class:
+
+* `void setupPeriodicFingerprintCleanup(boolean toggle, @CheckForNull int recurrence_period)`
```
afalko, line 129: recurrence_period should be long, also have users pass `TimeUnit` as a separate parameter.

afalko: or even better, use `Duration`: https://github.com/a0x8o/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L65

stellargo:
@oleg-nenashev updated the PR with the design we went with :) It is merge ready now

PR #301, opened 2020-09-15
Un-break the format

Wadeck:
@jglick


oleg-nenashev (approved):

jglick:
also done in #300 FYI

PR #295, opened 2020-07-22
Spring Security upgrade

jglick:
Description of a major migration centered around https://github.com/jenkinsci/jenkins/pull/4848.

"Submitted for approval as draft JEP by @jenkinsci/jep-editors.

PR #308, opened 2020-09-29
Removing jenkins-cloudfoundry-uaa

jglick:
In https://github.com/jenkinsci/jep/pull/295#discussion_r495468593 @daniel-beck points out that no such plugin exists on the update center.
Seems to have been a CloudBees proprietary plugin.


oleg-nenashev (approved):
Yes, I remember this plugin.

PR #310, opened 2020-10-01
[INFRA-1936] Switch JEP-217 to active

olblak:
https://issues.jenkins-ci.org/browse/INFRA-1936

### DIFF
--- a/jep/217/README.adoc
+++ b/jep/217/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
-| Draft :speech_balloon:
+| Active
 
 | Type
 | Standards


oleg-nenashev:
I do not think we've finished here. There is still an open comment from @bitwiseman about the organization naming. My preference would be to keep it open until we have a solution for the recent DockerHub policy changes. I can totally imagine that we may have to get rid of this experimental org after that

timja:
@olblak / @oleg-nenashev do we want this?

I don't think the policy changes really affect this, we aren't publishing much to jenkins4eval currently. I think some repositories publish there on the master branch on ci.jenkins.io to make it easier to test publishing scripts, but no users will be consuming it currently as far as I know.

olblak:
> no users will be consuming it currently as far as I know
And no users should be consuming images from there. We also had some discussion about deleting published images on a regular basis.
IMHO we can close this

PR #313, opened 2020-10-06
Assorted JEP-227 updates

jglick:


Wadeck (approved):
🐝

jep/227/compatibility.adoc
```diff
@@ -123,10 +123,9 @@ could be switched to `hasPermission` for full compatibility.
 |Test cleanup pending link:https://github.com/jenkinsci/extensible-choice-parameter-plugin/pull/42[extensible-choice-parameter-plugin #4].
 
 |link:https://plugins.jenkins.io/github-oauth/[github-oauth]
-|Incompatible, fixes prepared
+|Probably compatible.
 |link:https://github.com/jenkinsci/github-oauth-plugin/pull/119[github-oauth-plugin #119]
-is necessary.
-Given that, it seems to work.
+cleans up deprecated usages.
```
jglick, line 128: merge conflict w/ #314

jep/227/README.adoc
```diff
@@ -210,8 +209,9 @@ Exception Details:
 
 === Providing binary compatibility for all plugins implementing `SecurityRealm`
 
-Some plugins like `ldap` which implement `SecurityRealm` pose a special problem.
+Some plugins like `sfee` which implement `SecurityRealm` pose a special problem.
```
MarkEWaite, line 212: The `sfee` plugin is removed from distribution.  The stats data shows only 6 installations, 1.301 and earlier.  Is there a plugin example that poses this problem that is not removed from distribution? 

If not, it may be worth noting the state of that plugin.
```suggestion
Some plugins like `sfee` which implement `SecurityRealm` pose a special problem.
```

MarkEWaite (commented):
Is there a different example than the `sfee` plugin for a plugin that implements `SecurityRealm`?

jep/227/README.adoc
jglick: This particular plugin may not be important, it was just an example of source code that is particularly problematic to migrate.

MarkEWaite: I was unable to find the source code for that plugin in my searches.  Wouldn't it be better to use an example plugin where source code is readily available?

MarkEWaite (approved):

oleg-nenashev (changes requested):
Please resolve the merge conflict

jep/227/README.adoc
jglick: https://github.com/jenkinsci/sfee-plugin

oleg-nenashev (approved):
Thanks!

PR #315, opened 2020-10-14
Continuous Delivery of Jenkins Components

jglick:
Submitted for approval as draft JEP by @jenkinsci/jep-editors.


jep/0000/README.adoc
```diff
@@ -0,0 +1,243 @@
...
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+Maintainers of Jenkins component repositories on GitHub, especially plugins, can opt into continuous delivery (CD).
+In this mode, every successful build of the `master` branch by ci.jenkins.io results in a new release of the component.
+GitHub Actions are used to rebuild the component and deploy it to Artifactory,
```
halkeye, line 69: it feels weird to use a "competitors" product to build our product? is there any reason ci.jenkins can't do it?

jglick: Yes, as described in this document.

PierreBtz (commented):
I like the idea, a couple of remarks/questions:
* what about the load on the Artifactory server? With this process in place the number of release will be equals to the number of merged PRs. Isn't there a risk given there are already storage limitation AFAIU?
* about the artifact versions, IIUC it will be the same principle as the JEP-305, so something based on the git commit id. Meaning that if I try to understand which version an instance is using I might have a hard time figuring out what it contains without digging through the git commit history. I understand that the idea is to continuously deliver, yet plugins would still be bound to the Core which itself is not continuously delivered (or is it?).
* an intermediate solution might be to trigger the GH action on demand. This would also be useful for backports.

jep/0000/README.adoc
```diff
@@ -0,0 +1,243 @@
...
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+Maintainers of Jenkins component repositories on GitHub, especially plugins, can opt into continuous delivery (CD).
+In this mode, every successful build of the `master` branch by ci.jenkins.io results in a new release of the component.
```
dwnusbaum, line 68: I think this release model has obvious benefits in terms of simplicity and ease of use for Jenkins developers, but to me it also seems relatively undesirable for Jenkins administrators and users, especially for plugins using tools like Dependabot. Many changes to plugins are not user-facing, so I think it would be frustrating as a Jenkins administrator to be frequently prompted to update plugins when there are no meaningful or user-visible changes, especially since updating a plugin requires restarting Jenkins.

jglick:
> load on the Artifactory server

Indeed that is something to consider: implementing CD means encouraging lots of releases, thus bigger release repositories, so we need to be sure infrastructure is ready for that if this gets broad adoption. Fortunately there is no time pressure: we can freely experiment on a small scale and see how things hold up.

A related concern which I ought to discuss in the JEP is the risk that frequent releases of low-level components (POMs, API plugins, etc.) could cause cascading “Dependabot storms” like the notorious recent `junit:4.13.1`, though you can tune the frequency at which Dependabot checks for updates as well as throttle the number of outstanding PRs etc.

> if I try to understand which version an instance is using I might have a hard time figuring out what it contains without digging through the git commit history

I do not think anything significant changes in this regard. The proposal includes Release Drafter integration so the **Releases** page lists (typically) PR titles in a given release; the difference is that typically there would be just one entry per “release”, rather than a bunch clumped together. The version string will start with an integer which increases over time; an update from 130.xxxx → 137.xxxx is not intrinsically more or less meaningful than an update from 1.3 → 1.7.

> trigger the GH action on demand

Agreed, this is a good option to prototype as well—I think GH added this feature just recently.

> it would be frustrating as a Jenkins administrator to be frequently prompted to update plugins when there are no meaningful or user-visible changes

This is certainly a concern. You sometimes get these already, because we often have to cut plugin releases just to make PCT happy, but the situation would become more common. I wonder how much of this is an intrinsic problem vs. UX: can we improve the **Updates** tab GUI to offer a summary of plugin release notes, for example, or should the default UC check interval be made more than one day? We need to balance this concern, though, by contrasting to the current situation in which it is common for a PR with a minor feature or bug fix to be developed, reviewed, approved, merged, …and then left unreleased for days or weeks or months.

jep/0000/README.adoc
```diff
@@ -0,0 +1,243 @@
...
+merely pushing a commit to `master` (including merging a pull request)
+suffices to trigger a deployment,
+and the version number will uniquely and securely identify that commit,
+with no need to create a redundant Git tag.
+Component maintainers who wish to follow SemVer principles,
+encoding some semantics into version numbers,
+can still do so by appending the generated number as a “micro” component to a manually maintained `major.minor.` prefix.
+
+GitHub Actions are attractive in this context because they define a trust boundary naturally scoped to the repository:
+a given bot token is defined in only repository, useful in only that repository, and used only for a containerized build of that repository.
```
jvz, line 173: Such a detail can be abstracted through a secrets manager like Vault (or Azure Key Vault like in the release infra) for generating deploy keys for each run (as a neat side-effect, we could also sign the artifacts on demand using repo-specific certificates).

Note that this abstraction works for both the GitHub Actions workflow as well as a Jenkins workflow.

jvz (approved):
Sounds pretty good! I left a bit of an over-engineered suggestion around deployment key management.

jep/0000/README.adoc
jglick: What kind of “deploy keys” does Artifactory support beyond what is already outlined in this document? And what do you mean by “each run”? I do not think I am following your suggestion.

jvz: As in a temporary deploy token for each run of the action.

jglick: I do not see any practical way to accomplish that. The token would have to be generated on demand while the action runs by some privileged service which would have to establish the identity of the running action?

jvz: My idea is something like, suppose you're building `jenkinsci/foo`. Then the action or pipeline will request a deploy key for that particular project which has limited authorization to upload files to expected locations as before. If the build attempted to deploy code elsewhere, it would find itself unauthorized at this point. If you're planning to create tokens for each repo, that gets you halfway there. The dynamic secrets binding would help with credentials rotation, dynamic creation of tokens, and better authorization integration.

Do note that this idea I'm proposing is still likely overkill. I'm considering different methods of using an external secrets manager in an unrelated task, and I thought I'd share the idea in a JEP that's exploring a similar space.

jglick: > request a deploy key for that particular project

This is the problem. Request from some service TBD. And how does that service verify that the request is in fact coming from a GH Action run on that repository? https://github.community/t/verifying-connections-from-a-github-hosted-runner/17331 is the closest I can find, and basically unanswered. Can you send your `GITHUB_TOKEN` to that service and have it call back to the GH API and verify that the token has write access to the repo?

jvz: It's quite possible that GH Actions are too new to have integrations with Vault or other things.

jglick: Well I presume we would need to write any service ourselves anyway, since what it needs to do is generate Artifactory tokens with permissions driven by information in RPU. The question is whether there is some reliable way to have a GH Action run prove its authenticity to an external service.

If we find a way to do this, I do not dispute that it would be nicer than having the RPU build process save long-lived tokens to the repo.

jvz: Yeah, there would certainly need to be some glue somewhere. See also the GH Action for Vault: https://github.com/marketplace/actions/vault-secrets

jglick: I found that you can run

```bash
curl -s -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' https://api.github.com/installation/repositories | jq -r '.repositories[0].full_name'
```

to prove that a supplied token is coming from an app installed on the repository. I am however having trouble verifying that the app has write permissions to the repository; https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#get-the-authenticated-app is not available using this installation token (it requires a JWT). https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#get-an-app requires you to know the app id. I could try to call some other endpoint which requires write access to contents but this is kind of scary. The following seems to work:

```bash
TAG=permcheck-$RANDOM
curl -i -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -d '{"ref":"refs/tags/'$TAG'","sha":"'$GITHUB_SHA'"}' https://api.github.com/repos/$GITHUB_REPOSITORY/git/refs
curl -i -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -X DELETE https://api.github.com/repos/$GITHUB_REPOSITORY/git/refs/tags/$TAG
```

I will propose this system in the document.

daniel-beck: > The dynamic secrets binding would help with credentials rotation, dynamic creation of tokens, and better authorization integration.

Isn't the credential rotation more of a problem in a large, sprawling, heterogeneous environment? We have a single GH org and a known subset of repos, all with API accessible at all times.

While this approach looks interesting (and I've also considered the need for rotating secrets), the only advantage I can see is this is more likely to not run into untimely secret rotation problems. Or perhaps more likely, if actions are allowed to run in parallel.

jglick: Well my hope would be that the Artifactory token provisioning service would generate short-lived tokens, good for only one hour say, and each GH Action run would get a fresh token permitted only to deploy to the paths specified in RPU. I am pretty fuzzy on what is possible on the Artifactory side, though (and have no permissions to experiment).

daniel-beck: > each GH Action run would get a fresh token permitted only to deploy to the paths specified in RPU

That's the plan. There's some new stuff in Artifactory (time-limited access tokens) which would simplify this quite a bit—and could make a compelling case for a "credentials service". I'll need to check exactly what the requirements would be.

As a first step, we can store a credential in the GH secrets directly, and periodically update/rotate them from a locally executed script somewhere, without needing to operate a separate service.

jglick: > from a locally executed script

If this is safer than the current proposal of doing this automatically from RPU builds, then sure.

daniel-beck: > So far they don't seem to be possible credentials for Maven, just the API, which is a weird design choice for Artifactory.

PEBCAK: I have two users, used one's username and the other's API token 🤦 

So API tokens work. Now on to using expiring access tokens…

daniel-beck: https://repo.jenkins-ci.org/webapp/#/artifacts/browse/tree/General/snapshots/io/jenkins/plugins/solarized-theme/0.2-SNAPSHOT/solarized-theme-0.2-20201021.162950-2.hpi is a snapshot of a plugin I uploaded using a token that expired after an hour (and the user name is something I used when generating the token, there's no actual token by that name).

The problem right now seems that it's not possible to assign permissions from a "permission target" to an access token, just a group membership.

RPU will need a larger overhaul for this, I think, between generating groups, poking GitHub's secrets API to update the deployment secret valid for the next X hours (RPU run interval + 3 hours or so?), and changing the YAML file format to accommodate multi-module projects and the new automated builds.

jep/0000/README.adoc
```diff
@@ -0,0 +1,316 @@
...
+link:https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/[manual triggers]
+can be used to deploy from a given branch on demand, rather than automatically upon push.
+This is also likely to be the preferred trigger for backport branches.
+Compared to running MRP locally, this is still much less effort for maintainers,
+though such a trigger ought to be sure to validate that there is a passing Jenkins CI check before proceeding.
+
+=== Push vs. pull for Artifactory tokens
+
+Rather than having the RPU build push Artifactory tokens into repository secrets,
+which introduces some issues of token expiry and possible theft by repository owners,
```
daniel-beck, line 227: Expiring tokens allowed only to publish the appropriate artifacts can be pushed into repos periodically, so that would not be as much of a problem as this seems to imply.

While it would certainly be effort to fake being run in a GH action with an appropriate GITHUB_TOKEN, it's unclear to me why the validation measures described below offer additional protection than just pushing new tokens every hour or so that expire after a few hours. It seems that in both cases, users with legitimate access can obtain credentials that are still valid for a short while after losing said access.

While the "pull" model looks nicer (and would certainly involve fewer tokens being generated), I'd rather not have us operate a publicly accessible service that has Artifactory admin credentials to generate these access tokens. (It needs to be an admin account doing it so that we don't have to generate a ton of users too - only admins can generate access tokens with specific permissions through group memberships without real users behind them.)

Most of the benefit of this over JEP-221 seems to be in not requiring us to have additional infrastructure. Now this proposes a (simpler, but still) new service to build and operate. Between security concerns and our current overall infra situation, that seems like a worse solution.

jep/0000/README.adoc
```diff
@@ -0,0 +1,316 @@
...
+(Maybe also create an `ARTIFACTORY_USER` pseudo-secret to match.)
+
+== Motivation
+
+Traditionally, all Jenkins components have been released using the link:https://maven.apache.org/maven-release/maven-release-plugin/[Maven Release plugin] (MRP).
+This is typically invoked by a developer on a local clone of the repository in the `master` branch:
+
+[source,bash]
+----
+mvn -B release:prepare release:perform
```
daniel-beck, line 121:
```suggestion
mvn --batch-mode release:prepare release:perform
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,316 @@
...
+----
+
+It can then (with difficulty) verify that the App has write permission to the repository,
+as an Action token will (to prevent spoofing from low-privileged Apps):
+
+[source,bash]
+----
+tag=permcheck-$RANDOM
+curl -H "Authorization: Bearer $TOKEN" -d '{"ref":"refs/tags/'$tag'","sha":"'$sha'"}' https://api.github.com/repos/$repo/git/refs
+curl -H "Authorization: Bearer $TOKEN" -X DELETE https://api.github.com/repos/$repo/git/refs/tags/$tag
```
daniel-beck, line 250:
```suggestion
curl --header "Authorization: Bearer $TOKEN" --data '{"ref":"refs/tags/'$tag'","sha":"'$sha'"}' https://api.github.com/repos/$repo/git/refs
curl --header "Authorization: Bearer $TOKEN" --request DELETE https://api.github.com/repos/$repo/git/refs/tags/$tag
```

daniel-beck (commented):
I really like where this is going 🚀 

It seems all the discussions right now are about details. Adding some thoughts about those in comments below.

jep/0000/README.adoc
jglick: > changing the YAML file format to accommodate multi-module projects

Actually necessary? There is already a `github` field which can be used to aggregate YAML files pertaining to the same repo.

jep/0000/README.adoc
```diff
@@ -184,6 +190,73 @@ And on the flip side, the requirement for a secure execution environment is more
...
+(These are not blockers to experimentation on a few repositories.)
+
+Artifactory might not be able to handle the traffic.
+This is already a concern generally with our hosted Artifactory,
+but the Jenkins project is looking into what precisely the limits are.
+
+Jenkins administrators might tire of constantly seeing entries in the *Plugin Manager » Updates* tab.
+In many cases, there may be few or no behavioral changes in a release, just code cleanups or POM tidying.
+While having these releases is sometimes valuable for PCT, they are not valuable to administrators.
+One mitigation is to slow down the frequency at which the update center is automatically checked, currently one day.
```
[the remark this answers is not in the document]
jglick: > having the same major/minor version number

That would not work for plugins managed under this system unless they used the more complex version schemes with a partially hand-maintained component.

I will add your other suggestions.

jep/0000/README.adoc
```diff
@@ -0,0 +1,316 @@
...
+followed by PRs to higher-level components with their own releases.
+Excessive updates could consume a lot of CI time and exacerbate the previously mentioned risks.
+
+For any such issues, or for maintainers who prefer to do manual sanity checks prior to release rather than when merging PRs,
+there is another option:
+link:https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/[manual triggers]
+can be used to deploy from a given branch on demand, rather than automatically upon push.
+This is also likely to be the preferred trigger for backport branches.
+Compared to running MRP locally, this is still much less effort for maintainers,
+though such a trigger ought to be sure to validate that there is a passing Jenkins CI check before proceeding.
```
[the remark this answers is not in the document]
jglick: The latter, I think.

I do not really want the Action to be running tests, because that is just another environment to maintain, and Dockerized tests will be quietly skipped.

jep/0000/README.adoc
```diff
@@ -0,0 +1,316 @@
...
+Now knowing that it has been called from an App with write permissions,
+such as the deployment Action,
+it can create a new Artifactory token with a short expiry (say one hour)
+granted permission only to upload to the paths defined for this repository in RPU
+and return that token in its response.
+The Action would then bind this token to an environment variable for use from `settings.xml`.
+
+== Backwards Compatibility
+
+The link:https://github.com/jenkins-infra/update-center2[Jenkins update center generator] requires no modifications:
```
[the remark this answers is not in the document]
jglick: Will make a note to this effect, thanks.

jep/0000/README.adoc
jglick: All makes sense. I will update the document to clarify that we do not currently propose to take this approach.

jep/0000/README.adoc
```diff
@@ -0,0 +1,316 @@
...
+The GitHub runner is solely responsible for rebuilding binary artifacts (such as plugin `*.hpi`) from sources.
+This defends against certain supply-chain attacks:
+if ci.jenkins.io were compromised, at worst this could result in components with test failures being deployed.
+The deployed binaries would still have been built from the source files stored on GitHub.
+
+Currently we presume that component maintainers are not maliciously inserting backdoors into manually deployed binaries.
+So long as maintainers are granted direct access to Artifactory as well as the option to use CD, trusting them is unavoidable,
+and it is desirable to offer this option to maintainers in order for example to produce backport releases
+(since the current system only handles `master` pushes).
+If a maintainer were _not_ given Artifactory credentials (or simply not listed in RPU),
```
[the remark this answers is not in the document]
jglick: I am not sure I follow. What sort of “decisions about the component”? If using MRP, you must have both a valid Artifactory token _and_ write permission to the repository. (Maybe some malicious uses of a saved secret require owner permission rather than merely write?)

jep/0000/README.adoc
jglick: Since we are not leaning toward a pull-based approach, I am closing this thread and trying to capture outstanding design issues in the document.

jep/0000/README.adoc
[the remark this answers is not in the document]
daniel-beck: > What sort of “decisions about the component”?

Who else gets to be a maintainer. Whether to suspend distribution. How to configure the repo if they're not a maintainer. Who to invite as a committer.

As the YAML files with upload permissions are the best reference we have had for a long time for who gets such decisions for plugins, their meaning quickly grew beyond the simple, technical "allowed to upload".

As presumably there will always be someone who will have to make these decisions for a component unless it is abandoned, there will always be users listed in YAML files, even if all releases are done via GitHub.

I've discussed with @oleg-nenashev previously that we need to separate the technical ability to upload new releases from the "ownership" of a component (and perhaps define the latter better), but I so far it looks like neither of us found the time to do this.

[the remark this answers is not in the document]
jglick: OK, I think I follow what you are saying now, and will try to update the document to reflect that.

oleg-nenashev (approved):
Looks good from the structure standpoint. Happy to accept as draft. Maybe we should introduce a new category for Plugin developer infrastructure JEPs, e.g. 500. 3xx and 4xxx are dead categories unfortunately, so no strong opinion

https://github.com/jenkinsci/jep/blob/master/jep/221/README.adoc needs to be marked as superseded once merged. Is it correct @daniel-beck ?

jglick:
This is more like a 3xx than anything else, but I do not really care.

Yes this is intended to supersede JEP-221.

oleg-nenashev:
Let's keep rolling 2xx then . Merging with ID=229

jglick:
BTW I do not think you need to actually push to a PR branch in order to merge a JEP draft. Just `git pull --no-commit`, fix up numbers, commit, and push.

PR #327, opened 2020-11-05
reverse-proxy-auth-plugin #40 released

jglick:
https://github.com/jenkinsci/reverse-proxy-auth-plugin/pull/40


oleg-nenashev (approved):
"Probably compatible" sounds very reassuring, let's see how it flies. Thanks for the patch! 👍

jglick:
> very reassuring

Unfortunately the test coverage is very slim, so this is a best effort only.

oleg-nenashev (approved):

PR #332, opened 2020-11-09
[JEP-229] Updates

jglick:
Various updates, but most importantly describing the successfully prototyped integration with Release Drafter and usage of self-published Actions to avoid duplication.

### DIFF
--- a/jep/229/README.adoc
+++ b/jep/229/README.adoc
@@ -115,7 +115,7 @@ recognizes a flag in a repo’s YAML entry (`cd.enabled` set to `true`).
 If that flag is set, a group whose name is derived from the GitHub repository name is created in Artifactory.
 That group is granted the same permissions as regular maintainers, but contains no users.
 Instead, RPU periodically generates a temporary Artifactory access token that's a member of that group
-and saves it to the GitHub secrets `ARTIFACTORY_USERNAME` and `ARTIFACTORY_TOKEN` in the corresponding repository.
+and saves it to the GitHub secrets `MAVEN_USERNAME` and `MAVEN_TOKEN` in the corresponding repository.
 
 == Motivation
 
--- a/jep/229/README.adoc
+++ b/jep/229/README.adoc
@@ -88,14 +88,14 @@ after a link:https://ci.jenkins.io/job/Plugins/job/log-cli-plugin/job/master/21/
 triggered by a link:https://github.com/jenkinsci/log-cli-plugin/commit/934670f4a9b84f243f8c0d9bbfc7be30a242e32b[`master` push].
 The important parts of source setup are the link:https://github.com/jenkinsci/log-cli-plugin/blob/b37970d51e7c2d2d723f39fb7271a263f0d2083d/pom.xml#L11-L14[`version` definition]
 with a link:https://github.com/jenkinsci/log-cli-plugin/blob/b37970d51e7c2d2d723f39fb7271a263f0d2083d/.mvn/maven.config#L3[custom format]
-and a link:https://github.com/jenkinsci/log-cli-plugin/blob/d041abc4ff822401430b3bc00bf5da31f1d6e8d9/.github/workflows/release.yaml#L37-L49[GitHub Action definition]
-running link:https://github.com/jenkinsci/log-cli-plugin/blob/d041abc4ff822401430b3bc00bf5da31f1d6e8d9/.github/workflows/release.sh#L5[`mvn deploy`]
-with a link:https://github.com/jenkinsci/log-cli-plugin/blob/d041abc4ff822401430b3bc00bf5da31f1d6e8d9/.github/workflows/settings.xml[bot token for Artifactory].
+and a link:https://github.com/jenkinsci/log-cli-plugin/blob/b1ee6375b940508709ff64cd8e154ad99eb270ce/.github/workflows/cd.yaml[GitHub Action workflow]
+running link:https://github.com/jglick/jenkins-maven-cd-action/blob/3f1f2fa2739ce0747de9a2545aee5d54917a7abb/run.sh#L4[`mvn deploy`]
+with a link:https://github.com/jglick/jenkins-maven-cd-action/blob/3f1f2fa2739ce0747de9a2545aee5d54917a7abb/settings.xml[bot token for Artifactory].
 
 Note that the deployment skips tests and all other Maven mojos not necessary to produce artifacts, because the GitHub Action first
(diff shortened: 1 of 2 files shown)


jep/229/README.adoc
```diff
@@ -100,29 +100,24 @@ It is the responsibility of the
...
-The problem is that it is not possible to assign permissions from a “permission target” to an access token, just a group membership.
-RPU will need a larger overhaul for this, between generating/maintaining groups;
-poking GitHub’s Secrets API to update the deployment secret to be valid for some period (such as the RPU run interval + 3 hours or so);
-and perhaps changing the YAML file format to accommodate multi-module projects and the new automated builds.
+link:https://github.com/jenkins-infra/repository-permissions-updater[Repository Permissions Updater (RPU) processing]
+recognizes a flag in a repo’s YAML entry ('cd.enabled` set to `true`)
+and creates an Artifactory bot account with permissions to only the listed paths
+(via a synthetic Artifactory group),
+then generates an access token for that account
+and saves it to the GitHub secrets `ARTIFACTORY_USERNAME` and `ARTIFACTORY_TOKEN` in the corresponding repository.
```
daniel-beck, line 120: Groups are real (and otherwise unused), accounts are not.

```suggestion
recognizes a flag in a repo’s YAML entry ('cd.enabled` set to `true`).
If that flag is set, a group whose name is derived from the GitHub repository name is created in Artifactory.
That group is granted the same permissions as regular maintainers, but contains no users.
Instead, RPU periodically generates a temporary Artifactory access token that's a member of that group
and saves it to the GitHub secrets `ARTIFACTORY_USERNAME` and `ARTIFACTORY_TOKEN` in the corresponding repository.
```

jep/229/README.adoc
```diff
@@ -88,41 +88,34 @@ after a link:https://ci.jenkins.io/job/Plugins/job/log-cli-plugin/job/master/21/
...
-The problem is that it is not possible to assign permissions from a “permission target” to an access token, just a group membership.
-RPU will need a larger overhaul for this, between generating/maintaining groups;
-poking GitHub’s Secrets API to update the deployment secret to be valid for some period (such as the RPU run interval + 3 hours or so);
-and perhaps changing the YAML file format to accommodate multi-module projects and the new automated builds.
+link:https://github.com/jenkins-infra/repository-permissions-updater[Repository Permissions Updater (RPU) processing]
+recognizes a flag in a repo’s YAML entry (`cd.enabled` set to `true`).
+If that flag is set, a group whose name is derived from the GitHub repository name is created in Artifactory.
+That group is granted the same permissions as regular maintainers, but contains no users.
+Instead, RPU periodically generates a temporary Artifactory access token that's a member of that group
+and saves it to the GitHub secrets `ARTIFACTORY_USERNAME` and `ARTIFACTORY_TOKEN` in the corresponding repository.
```
daniel-beck, line 118: Would it make sense to be more implementation agnostic here and call these `MAVEN_REPO_USERNAME` / `MAVEN_REPO_TOKEN` instead?

jep/229/README.adoc
```diff
@@ -351,6 +367,14 @@ implements basic aspects of this proposal from the developer side.
 link:https://github.com/jenkins-infra/repository-permissions-updater/pull/1747[repository-permissions-updater #1747]
 implements changes to infrastructure.
 
+link:https://github.com/jenkinsci/plugin-pom/pull/375[plugin-pom #375] and link:https://github.com/jenkinsci/pom/pull/147[pom #147]
+add a POM profile to block MRP in CD mode.
```
daniel-beck, line 371: It is unclear what this is for. Prevent accidental misuse?

daniel-beck (approved):

jep/229/README.adoc
jglick: Fine with me. If you want to make such a change in RPU and also push fresh secrets with those names to `log-cli-plugin`, I will update everything to match.

jep/229/README.adoc
jglick: Yes, prevents you from running `mvn release:prepare release:perform`, which would produce crazy results. If you opt into `changelist.format`, you agree not to use MRP.

jep/229/README.adoc
```diff
@@ -88,41 +88,34 @@ after a link:https://ci.jenkins.io/job/Plugins/job/log-cli-plugin/job/master/21/
...
 
-As of this writing unimplemented:
-
-* Use the Action for Release Drafter to create a release (typically including exactly one entry, from a recently merged PR).
-* Create a new Action wrapping the `release.sh` + `settings.xml` so we do not have this stuff duplicated.
-* Add `block-MRP` profile to `plugin-pom`.
+Release Drafter is configured as an Action as well, but always publishing the draft release under the placeholder name `next` rather than a real version.
+This runs (fairly) quickly after PRs are merged, and just updates a draft release.
+If and when a release is successfully deployed to Artifactory, the deployment script
+link:https://github.com/jglick/jenkins-maven-cd-action/blob/3f1f2fa2739ce0747de9a2545aee5d54917a7abb/run.sh#L5-L8[creates a (lightweight) Git tag and publishes the release].
```
timja, line 106: is the lightweight bit relevant?

jglick: Well, somewhat; there is a different GitHub API for creating an annotated tag, which I am not doing.

jep/229/README.adoc
```diff
@@ -286,6 +279,29 @@ On balance this “pull” approach seems worse than the currently proposed “p
 * Theft of tokens prior to expiry by malicious maintainers is a possibility under either system;
   the window of opportunity would differ, as would the sort of audit trail produced.
 
+=== Release Drafter as a step in the release workflow
+
+Unfortunately it is not possible to run Release Drafter as a step in the release workflow,
```
oleg-nenashev, line 284: https://github.com/jenkinsci/core-changelog-generator has a changelog generator mode which takes the Release Drafter format. If needed, we can use it in the recommended flow inside Jenkins or elsewhere

oleg-nenashev: FTR https://github.com/jenkinsci/core-changelog-generator/pull/12 . Some patches are still needed to support all Release Drafter options, the patch was designed specifically for the Jenkins core

jglick: Thanks. I think it is safest to stick with using official RD even if it needs to be done as a separate workflow at least for now. Can be revisited as we get real experience though. Will note your idea.

timja (approved):

jglick:
ping @oleg-nenashev or any other JEP editors

oleg-nenashev (approved):

PR #334, opened 2020-11-10
Marking JEP-227 & JEP-228 accepted

jglick:
I do not think we have any working formal definition of “accepted” after the loss of a BDFL but having the core PRs released as 2.266 seems the closest we can come.

### DIFF
--- a/jep/227/README.adoc
+++ b/jep/227/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
-| Draft :speech_balloon:
+| Accepted :ok_hand:
 
 | Type
 | Standards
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
I suggest discussing it at the Governance Meeting. I also do not see any reason to go through the dead process, but we need an explicit consensus on that

daniel-beck (approved):

jglick:
> we need an explicit consensus on that

Whatever you prefer. I do not really care that much, just trying to make the documents reflect reality to the extent possible.

timja:
> . **<<accepted, Accepted Status>>** - Andrea and other contributors
  complete all remaining implementation related to the
  "<<accepted, Accepted>>" JEP (code, documentation, or other changes).

Final makes more sense to me but lets move this forward

PR #336, opened 2020-11-18
[JENKINS-65988] Upgrade guava to 30.0-jre

dbreheret:
As advised by Matt Sicker, created a jep for guava upgrade.

### DIFF
--- a/jep/311/README.adoc
+++ b/jep/311/README.adoc
@@ -75,7 +75,7 @@ much more.
  Guava (complete) has only one code dependency - javax.annotation,
     per the JSR-305 spec.
 
-See the https://xxxhttps://github.com/google/guava/releases/tag/v30.0[Guava release notes]
+See the link:https://github.com/google/guava/releases/tag/v30.0[Guava release notes]
 
 == Motivation
 
--- a/jep/311/README.adoc
+++ b/jep/311/README.adoc
@@ -1,0 +1,131 @@
+= JEP-311: Upgrade guava to 30.0-jre
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.**JEP Template**
+
+.Metadata
+[cols="1h,1"]
+|===
+| JEP
+| 311
+
+| Title
+| Upgrade guava to 30.0-jre
+
+| Sponsor
+| link:https://github.com/dbreheret[Dominique Breheret]
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Draft :speech_balloon:
+
+| Type
+| Standards
+
+| Created
+| 2020-11-18
+
+| BDFL-Delegate
+| TBD
+
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+//| JIRA
+//| :bulb: https://issues.jenkins-ci.org/browse/JENKINS-nnnnn[JENKINS-nnnnn] :bulb:
+//
+//
+// Uncomment if discussion will occur in forum other than jenkinsci-dev@ mailing list.
+//| Discussions-To
+//| :bulb: Link to where discussion and final status announcement will occur :bulb:
+
+| Requires
+|
+
+// Uncomment and fill if this JEP is rendered obsolete by a later JEP
+//| Superseded-By
+//| :bulb: JEP-NUMBER :bulb:
+//
+//
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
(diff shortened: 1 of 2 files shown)


jvz:
The main thing that may cause issues here will be in plugins. For example, searching for `com.google.common` in this org (note that Guice, Gson, and likely some other Google libraries may show up in this search, so you might have a better idea on what to look for than that package): https://github.com/search?q=org%3Ajenkinsci+filename%3A*.java+in%3Afile+com.google.common&type=code

In order to upgrade Guava, we'll likely need some additional glue code for backward compatibility with the version distributed in Jenkins if anything here breaks. The [Spring Security upgrade JEP](https://github.com/jenkinsci/jep/tree/master/jep/227) shows how this can potentially be non-trivial. These concerns are what you can describe in the "Compatibility" section of this JEP.

jep/311/README.adoc
```diff
@@ -0,0 +1,131 @@
...
+Besides tests inside Jenkins core itself,
+CloudBees will endeavor to verify that all
+link:https://docs.cloudbees.com/search?&type=ci-plugins&ci-plugins-tier=verified[“Tier 1”] and
+link:https://docs.cloudbees.com/search?&type=ci-plugins&ci-plugins-tier=compatible[“Tier 2”]
+plugins are compatible with the core changes,
+as determined by acceptance tests (ATH) and `plugin-compat-tester` (PCT).
+
+== Prototype Implementation
+
+link:https://github.com/jenkinsci/jenkins/pull/5059[jenkins #5059] is the main change.
```
jglick, line 126: https://github.com/jenkinsci/jenkins/pull/5059

jep/311/compatibility.adoc
```diff
@@ -0,0 +1,176 @@
+= JEP-311: Upgrading guava to 30.0-jre
+
+Use this space to track the status of plugins that may or may not be compatible with JEP-311.
+Refer to general information about link:README.adoc#backwards-compatibility[backwards compatibility]
+for tips on searching for potentially problematic API usages.
+
+Plugins which do not do anything special with XStream need not be listed.
```
jglick, line 7: copy-pasta
```suggestion
Plugins which do not do anything special with Guava need not be listed.
```

jep/311/compatibility.adoc
```diff
@@ -0,0 +1,176 @@
...
+|
+
+|link:https://plugins.jenkins.io/allure-jenkins-plugin/[allure-jenkins-plugin]
+|To be investigated
+|
+
+|link:https://plugins.jenkins.io/ant/[ant]
+|To be investigated
+|
+
```
jglick, line 31: Also: https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/83

jep/311/README.adoc
```diff
@@ -0,0 +1,131 @@
...
+// Uncomment when this JEP status is set to Accepted, Rejected or Withdrawn.
+//| Resolution
+//| :bulb: Link to relevant post in the jenkinsci-dev@ mailing list archives :bulb:
+
+|===
+
+== Abstract
+
+Jenkins uses guava version 11.0.1, which is seriously dated.
+Security AppScan are not happy about this version.
```
jvz, line 67: Note that security scans detect a serialization-related vulnerability in Guava, but Jenkins uses [JEP 200](https://github.com/jenkinsci/jep/tree/master/jep/200) to form an explicit list of allowed classes for deserialization, and the affected Guava class(es) are not and will never be added to the list. Getting security scanners to stop complaining about it, though, is certainly a useful goal as that was why there were recent JEPs to do all the tedious work to upgrade both Spring Security and XStream, both of which also had security scan complaints that were irrelevant in Jenkins and involved libraries that had broken backward compatibility.

However, if we can successfully upgrade Guava, then that will make it so that plugins might actually be able to more reliably use Guava themselves rather than avoiding it due to being an outdated dependency.

jep/311/README.adoc
```diff
@@ -0,0 +1,131 @@
+= JEP-311: Upgrade guava to 30.0-jre
```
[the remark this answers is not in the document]
oleg-nenashev: I can change it before merging

oleg-nenashev (changes requested):
Thanks for starting this draft @dbreheret!

Requesting changes because of the CloudBees commitment reference. If there is no explicit and public commitment from a company contributor, it should not be mentioned like that even in the draft.

For other matters, there is a pending conversation in https://github.com/jenkinsci/jenkins/pull/5059 which could be taken into account in the JEP specification. Addressing this feedback does not block the JEP from being merged as a draft

olamy:
BTW we are now guava 30.1-jre :)

dbreheret:
Smashing :)

oleg-nenashev:
@jenkinsci/core @basil @batmat Taking https://groups.google.com/g/jenkinsci-dev/c/aYUJ4VuOuVc/m/I9IEEzovAgAJ , would we like to revive and continue this proposal? CC @dbreheret in case he is still interested

dbreheret:
> 
> 
> @jenkinsci/core @basil @batmat Taking https://groups.google.com/g/jenkinsci-dev/c/aYUJ4VuOuVc/m/I9IEEzovAgAJ , would we like to revive and continue this proposal? CC @dbreheret in case he is still interested

Yes, still interested.

jtnord:
obsoleted by #375?

PR #342, opened 2020-12-02
[JEP-228] Noting allure release

jglick:
https://github.com/jenkinsci/allure-plugin/pull/271 @eroshenkoam


eroshenkoam:
@jglick ty, sorry for such delay

jglick:
> such delay

A few weeks you mean? That is a prompt response compared to what I usually deal with. :grin: Just happy to see a fix released.

oleg-nenashev (approved):

PR #345, opened 2021-02-24
JEP-15 - Update to LFX Crowdfunding

oleg-nenashev:


MarkEWaite (approved):
Looks good.  I checked several links and confirmed that they work as expected.

PR #364, opened 2021-06-01
Also remove `stapler-jruby`

jglick:
https://github.com/stapler/stapler/pull/238#issuecomment-852409892


daniel-beck (approved):
🎉 

I thought I had mentioned the cleanup we can do there, but that probably was in private chat somewhere.

oleg-nenashev (approved):

PR #373, opened 2021-07-20
Maybe move `jenkins.slaves.restarter`

jglick:
https://github.com/jenkinsci/remoting-opentelemetry-plugin/issues/65#issuecomment-883754559 for context. I need to think about this a bit more.

### DIFF
--- a/jep/230/README.adoc
+++ b/jep/230/README.adoc
@@ -88,6 +88,8 @@ split into a new plugin such as `pipeline-libraries` without this dependency.
 The `jenkins-module` packaging could be removed from `maven-hpi-plugin`, if there are no other users.
 (Theoretically some vendors could be using it for a `ConfidentialStore` implementation, for example.)
 
+The core classes in the `jenkins.slaves.restarter` package could be moved to the `slave-installer` plugin.
+
 == Motivation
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
CC @Aki-7 for the reference

oleg-nenashev (commented):
I am not sure about it on my own. The Agent installer components were designed for installing the payload on the agent side, not for other common operations like restarts. While we could change the design of the components, it does not seem to fit the existing architecture

timja:
FTR the restarters are broken on Mac M1 ARM

Agent throws a class not found exception and controller exits without restarting

jglick:
> The Agent installer components were designed for installing the payload on the agent side, not for other common operations like restarts.

Right; my reasoning was that attempting to have an agent restart in place (rather than just exit and let a service launch a new copy) falls into the same category of stuff that only makes sense for agents launched interactively via JWS that can be omitted from modern Jenkins installations.

jglick:
I would need to spend more time thinking through implications and testing before actually doing it. Approving an edit to the JEP does not imply approving the _action_; the whole thing remains in draft anyway (despite SSHD being done).

timja (approved):

PR #375, opened 2021-09-05
Upgrade Guava from 11.0.1 to latest

basil:
Description of a major migration centered around jenkinsci/jenkins#5707 and [JENKINS-65988](https://issues.jenkins.io/browse/JENKINS-65988).

Submitted for approval as draft JEP by @jenkinsci/jep-editors.

PR #376, opened 2021-10-06
Suggesting automatable procedure for merging JEP drafts

jglick:
Formalizing https://github.com/jenkinsci/jep/pull/375#issuecomment-929184109. I do not like pushing to other people’s PRs, and in the past we have sometimes accidentally mangled JEP PRs trying to be too clever. This procedure works, is straightforward, and retains history of what the JEP submitter wrote vs. what the JEP editor did to merge.

### DIFF
--- a/jep/1/README.adoc
+++ b/jep/1/README.adoc
@@ -918,15 +918,28 @@ revision with specific instructions.
 
 ==== Approve as Draft
 
-Once the JEP is ready for the repository, a JEP editor will:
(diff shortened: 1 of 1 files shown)


jglick:
@MarkEWaite as in https://github.com/jenkinsci/jep/pull/380#issuecomment-976845370 perhaps all usages of `origin` (incl. the implicit one in `git pull`) should be replaced with `https://github.com/jenkinsci/jep`?

PR #379, opened 2021-11-03
[JEP-228] Noting `envinject-api` fix

jglick:
Follows up #378 to note https://github.com/jenkinsci/envinject-api-plugin/pull/17.


basil (approved):

oleg-nenashev (approved):
Thanks for thé fix!

PR #393, opened 2022-09-19
JEP submission/maven repo 2022

dduportal:
This PR is a proposal around https://github.com/jenkins-infra/helpdesk/issues/2322.

Its goal is to describe and propose a solution for the missions of repo.jenkins-ci.org in regards to its sustainability.

WiP...

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -55,6 +55,14 @@ Realign the missions and goals of repo.jenkins-ci.org (JFrog SaaS-Hosted's Artif
 
 == Specification
 
+=== TL;DR;
+
+- Enable authentication on the repo.jenkins-ci.org mirrored repositories to limit who is able to download through our mirros
+- Keep hosted repositories unauthenticated to keep a low contributing initial entry bar when building the Jenkins projects
+- Add caching for our infrastructure build with full authenticated process
+
+=== Proposal
+
 [TIP]
 ====
 Provide a detailed specification of what is being proposed.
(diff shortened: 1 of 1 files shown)


basil (commented):
My understanding is that the JFrog bandwidth project has come to a close, so can this PR be closed?

samrocketman:
This proposal would be detrimental to how I operate.  I have my own artifact store and have the Jenkins maven repo as a proxy.  I build out images using caching through Nexus.  I hope there's not ever plans to implement something like this.

MarkEWaite:
> This proposal would be detrimental to how I operate. I have my own artifact store and have the Jenkins maven repo as a proxy. I build out images using caching through Nexus. I hope there's not ever plans to implement something like this without some kind of maven compatible alternative.

I don't think there is much chance of us ever implementing something that is not maven compatible.  Maven is at the center of all our build processes.

I don't think there is any need to implement something that is not maven compatible.  JFrog is happy to host the artifacts that the Jenkins project produces, like the plugins and war files.  Jenkins developers rely on the Maven repository provided by the Jenkins project and JFrog continues to provide that repository.

JFrog asked that we reduce bandwidth use from that repository.  We analyzed the log files and found that there were instances of misuse.  We blocked some of those instances of misuse.  We persuaded others to stop their misuse.  We stopped caching the Maven central repository to further reduce bandwidth use.  Those steps were sufficient for JFrog and did not disrupt Jenkins development.

I agree with @basil that this pull request can be closed.

samrocketman:
I see; thank you for additional context.  I realize now that date of this is PR old as well.

dduportal:
Absolutely, we can close as https://github.com/jenkins-infra/helpdesk/issues/3599 was implemented and confirmed validated by JFrog. Sorry that I left this PR outside my radar.

PR #398, opened 2023-09-06
FIPS-140 Jep

jtnord:
new Jep for FIPS-140 compliance.

### Testing done

I checked the rendering locally before creating the PR and it looked ok.

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -19,7 +19,7 @@ endif::[]
 | FIPS-140 Support
 
 | Sponsor
-| link:https://github.com/jtnord
+| link:https://github.com/jtnord[James Nord]
 
 // Use the script `set-jep-status <jep-number> <status>` to update the status.
 | Status
@@ -29,7 +29,7 @@ endif::[]
 | Standards
 
 | Created
-| Date (2023-09-06)
+| 2023-09-06
 
 | BDFL-Delegate
 | TBD
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -94,7 +94,7 @@ As of the time of writing the validated Java implementations are only FIPS-140-2
 
 It is out of scope to support previous configuration, as it would not be possible to migrate non FIPS-140 compliant data to FIPS-140 compliant data within a FIPS-140 compliant environment.  As such there is no backwards compatibility with previous versions.
 
-It may but is not a recquirement of this JEP be possible to switch from a FIPS-140 compliant environment to a non FIPS-140 compliant environment.
+It may but is not a requirement of this JEP be possible to switch from a FIPS-140 compliant environment to a non FIPS-140 compliant environment.
 
 Code changes will however be backwards compatible; if the system is running in a non FIPS-140 compliant environment then it will continue to run and operate the same after any changes due to this JEP.
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -82,7 +82,7 @@ Configuration of the JVM, platform etc. is out-of-scope, as is making every feat
 Currently it is not possible to run Jenkins in a link:https://csrc.nist.gov/pubs/fips/140-2/upd2/final[FIPS-140-2] compliant way.  Many users needing Jenkins are required (e.g. for US government projects) to have any solution be compliant with FIPS-140 and today it is not possible for them to do this. 
 
 Both Jenkins core and plugins use cryptography that is not allowed and/or are not from an approved implementation, and this prevents adoption of Jenkins by these end users.
-Examples would be during setup where the install wizard creates a first admin user using the internal security realm which uses ``bcrypt`` which would fall fowl of the standard due to its use of ``blowfish`` which is not an approved algorithm.
+Examples would be during setup where the install wizard creates a first admin user using the internal security realm which uses ``bcrypt`` which would fall foul of the standard due to its use of ``blowfish`` which is not an approved algorithm.
 
 == Reasoning
 
@@ -100,7 +100,7 @@ Code changes will however be backwards compatible; if the system is running in a
 
 == Security
 
-Some of the cryptography supported in FIPS 140 (especially FIPS 140-2) is considered less secure than some modern alternatives.  There is no alternative in cases like these where FIPS-140 compliance is required.
+Some of the cryptography supported in FIPS 140 (especially FIPS 140-2) is considered less secure than some modern alternatives.  FIPS 140-2 cryptography will be used when Jenkins is running in FIPS-140 mode even if that cryptography is considered less secure than other alternatives.
 
 == Infrastructure Requirements
 
@@ -108,9 +108,9 @@ There are no new infrastructure requirements related to this proposal.
 
 == Testing
 
-New code that changes behaviour when it detects the desire to run in FIPS-140 mode would be covered by unit tests like any other code change.  There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity and code of maintaining this setup.
+New code that changes behaviour when it detects the desire to run in FIPS-140 mode would be covered by unit tests like any other code change.  There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity of maintaining this setup.
 
-It is possible that regressions could be made due to the lack of automation to catch issues, however automation can only catch pre-known issues, and could not detected for example the introduction of a new encryption routine from a previously unknown library.
+Test automation is not expected to detect code in Jenkins core or in Jenkins plugins that is not FIPS-140 compliant.
 
 == Prototype Implementation
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -86,7 +86,10 @@ Examples would be during setup where the install wizard creates a first admin us
 
 == Reasoning
 
-In order to be FIPS-140 compliant the encryption algorithms used need to be validated, this requires complex configuration of the host, the JVM as well as the changes to Jenkins and plugin code as well as checking of any plugin code that would be installed.  As such it is not considered that the Jenkins community supports, or contains instructions on how to run Jenkins in a compliant manner, but merely contains only an overview to say it may be possible with various configuration of the JVM, Servlet environment, limitations of plugins and finally instructing Jenkins to prefer a FIPS-140 mode of operation.
+To be FIPS-140 compliant, Jenkins must use validated, FIPS-140 compliant encryption algorithms. 
+This requires complex configuration of the host, the Java Virtual Machine (JVM), the servlet environment, and the Jenkins controller. 
+Changes may be required in both Jenkins core and plugins.
+As such it is not considered that the Jenkins community supports or provides instructions on how to run Jenkins in a compliant manner, but merely states that FIPS-140 compliant operation may be  possible with appropriate configuration of the JVM, Servlet environment, limited set of plugins, and finally instructing Jenkins to prefer a FIPS-140 mode of operation.
 
 As of the time of writing the validated Java implementations are only FIPS-140-2 compliant even though FIPS-140-3 has superseded it.  As such the initial implementations will be targeting the older FIPS-140-2 standard.
 
--- a/jep/0000/README.adoc
+++ b/jep/0000/README.adoc
@@ -1,0 +1,124 @@
+= JEP-0000: FIPS-140 support
+:toc: preamble
+:toclevels: 3
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="1h,1"]
+|===
+| JEP
+| 0000
+
+| Title
+| FIPS-140 Support
+
+| Sponsor
+| link:https://github.com/jtnord
+
+// Use the script `set-jep-status <jep-number> <status>` to update the status.
+| Status
+| Not Submitted :information_source:
+
+| Type
+| Standards
+
+| Created
+| Date (2023-09-06)
+
+| BDFL-Delegate
+| TBD
+
+//
+//
+// Uncomment if there is an associated placeholder JIRA issue.
+// the following is just one Jira - there may be more so not spcifically linking
+//| JIRA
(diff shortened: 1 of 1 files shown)


jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+As of the time of writing the validated Java implementations are only FIPS-140-2 compliant even though FIPS-140-3 has superseded it.  As such the initial implementations will be targeting the older FIPS-140-2 standard.
+
+== Backwards Compatibility
+
+It is out of scope to support previous configuration, as it would not be possible to migrate non FIPS-140 compliant data to FIPS-140 compliant data within a FIPS-140 compliant environment.  As such there is no backwards compatibility with previous versions.
+
+It may but is not a recquirement of this JEP be possible to switch from a FIPS-140 compliant environment to a non FIPS-140 compliant environment.
```
timja, line 97:
```suggestion
It may but is not a requirement of this JEP be possible to switch from a FIPS-140 compliant environment to a non FIPS-140 compliant environment.
```

olamy (approved):

jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+* A new API will be introduced that code can use to check if FIPS-140 compliance is preferred.
+* Code in Jenkins core (where necessary) switches behaviour between more secure yet non compliant cryptography to FIPS-140 compliant cryptography.
+* Non compliant and less secure crytpography has been replaced by a secure and compliant option.
+* General guidelines exist to help developers write FIPS-140 compliant code and/or understand incompatibilities with the standard.
+* User facing documentation provides a mere overview of the FIPS flag and what it does (and does not do).
+
+== Motivation
+
+Currently it is not possible to run Jenkins in a link:https://csrc.nist.gov/pubs/fips/140-2/upd2/final[FIPS-140-2] compliant way.  Many users needing Jenkins are required (e.g. for US government projects) to have any solution be compliant with FIPS-140 and today it is not possible for them to do this.
```
samrocketman, line 82:
```suggestion
Currently it is not possible to run Jenkins in a link:https://csrc.nist.gov/pubs/fips/140-3/final[FIPS-140-3] compliant way.  Many users needing Jenkins are required (e.g. for US government projects) to have any solution be compliant with FIPS-140 and today it is not possible for them to do this. 
```

FIPS-140-2 page (published 2001) you link says it is superseded by FIPS-140-3 (published 2019).

samrocketman: Never mind; under reasoning the following is listed.

> As of the time of writing the validated Java implementations are only FIPS-140-2 compliant even though FIPS-140-3 has superseded it. As such the initial implementations will be targeting the older FIPS-140-2 standard.

rsandell (approved):

jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+* Non compliant and less secure crytpography has been replaced by a secure and compliant option.
+* General guidelines exist to help developers write FIPS-140 compliant code and/or understand incompatibilities with the standard.
+* User facing documentation provides a mere overview of the FIPS flag and what it does (and does not do).
+
+== Motivation
+
+Currently it is not possible to run Jenkins in a link:https://csrc.nist.gov/pubs/fips/140-2/upd2/final[FIPS-140-2] compliant way.  Many users needing Jenkins are required (e.g. for US government projects) to have any solution be compliant with FIPS-140 and today it is not possible for them to do this. 
+
+Both Jenkins core and plugins use cryptography that is not allowed and/or are not from an approved implementation, and this prevents adoption of Jenkins by these end users.
+Examples would be during setup where the install wizard creates a first admin user using the internal security realm which uses ``bcrypt`` which would fall fowl of the standard due to its use of ``blowfish`` which is not an approved algorithm.
```
MarkEWaite, line 85: Minor spelling error, at least according to the [Merriam-Webster dictionary](https://www.merriam-webster.com/dictionary/fall%20foul%20of).  Non-blocking change.
```suggestion
Examples would be during setup where the install wizard creates a first admin user using the internal security realm which uses ``bcrypt`` which would fall foul of the standard due to its use of ``blowfish`` which is not an approved algorithm.
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+== Reasoning
+
+In order to be FIPS-140 compliant the encryption algorithms used need to be validated, this requires complex configuration of the host, the JVM as well as the changes to Jenkins and plugin code as well as checking of any plugin code that would be installed.  As such it is not considered that the Jenkins community supports, or contains instructions on how to run Jenkins in a compliant manner, but merely contains only an overview to say it may be possible with various configuration of the JVM, Servlet environment, limitations of plugins and finally instructing Jenkins to prefer a FIPS-140 mode of operation.
```
MarkEWaite, line 89: The sentence seems like it contains multiple concepts that could be split into separate sentences.  Google Bard (and some edits of my own) suggested the following sentences.  Optional feedback that may use the wrong words or the wrong concepts.  You're welcome to ignore the comment
```suggestion
To be FIPS-140 compliant, Jenkins must use validated, FIPS-140 compliant encryption algorithms. This requires complex configuration of the host, the Java Virtual Machine (JVM), the servlet environment, and the Jenkins controller. It requires changes to Jenkins core and to Jenkins plugins.. As such it is not considered that the Jenkins community supports or provides instructions on how to run Jenkins in a compliant manner, but merely states that FIPS-140 compliant operation may be  possible with appropriate configuration of the JVM, Servlet environment, limited set of plugins, and finally instructing Jenkins to prefer a FIPS-140 mode of operation.
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+It is out of scope to support previous configuration, as it would not be possible to migrate non FIPS-140 compliant data to FIPS-140 compliant data within a FIPS-140 compliant environment.  As such there is no backwards compatibility with previous versions.
+
+It may but is not a requirement of this JEP be possible to switch from a FIPS-140 compliant environment to a non FIPS-140 compliant environment.
+
+Code changes will however be backwards compatible; if the system is running in a non FIPS-140 compliant environment then it will continue to run and operate the same after any changes due to this JEP.
+
+== Security
+
+Some of the cryptography supported in FIPS 140 (especially FIPS 140-2) is considered less secure than some modern alternatives.  There is no alternative in cases like these where FIPS-140 compliance is required.
```
MarkEWaite, line 103: Optional rephrasing of the second sentence.  You're welcome to ignore this rephrasing as well.  The existing phrasing states the facts correctly.
```suggestion
Some of the cryptography supported in FIPS 140 (especially FIPS 140-2) is considered less secure than some modern alternatives.  FIPS 140-2 cryptography will be used when Jenkins is running in FIPS-140 mode even if that cryptography is considered less secure than other alternatives.
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+== Infrastructure Requirements
+
+There are no new infrastructure requirements related to this proposal.
+
+== Testing
+
+New code that changes behaviour when it detects the desire to run in FIPS-140 mode would be covered by unit tests like any other code change.  There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity and code of maintaining this setup.
```
MarkEWaite, line 111: The "and code" seems redundant here.  Optional.
```suggestion
New code that changes behaviour when it detects the desire to run in FIPS-140 mode would be covered by unit tests like any other code change.  There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity of maintaining this setup.
```

jep/0000/README.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+It is possible that regressions could be made due to the lack of automation to catch issues, however automation can only catch pre-known issues, and could not detected for example the introduction of a new encryption routine from a previously unknown library.
```
MarkEWaite, line 113: This seems like a general statement of the "imperfect but useful" state of test automation.  I'd remove it (first preference) or rephrase it significantly (suggestion below).
```suggestion
Test automation is not expected to detect code in Jenkins core or in Jenkins plugins that is not FIPS-140 compliant.
```

MarkEWaite (approved):
Optional comments for your consideration.  Approved whether or not those comments are included in the proposal.

jep/0000/README.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+== Testing
+
+New code that changes behaviour when it detects the desire to run in FIPS-140 mode would be covered by unit tests like any other code change.  There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity of maintaining this setup.
+
+Test automation is not expected to detect code in Jenkins core or in Jenkins plugins that is not FIPS-140 compliant.
+
+== Prototype Implementation
+
+link:https://github.com/jenkinsci/jenkins/pull/8393[Jenkins#8393] contains a prototype for enabling a FIPS-140 preference flag that code can adapt behaviour to, as well as a change in behaviour for the ``HudsonPrivateSecurityRealm``.
```
basil, line 120: OK, but this seems like it just scratches the surface in terms of exploring the relevant issues.

jtnord: where you thinking of anything in particular?  

We have audited the Jenkins codebase (still going through several libraries that it exposes to plugins and uses) but the code in jenkinsci/jenkins should now all be ok.

FIPS-140 only cares about things in a security context.  For an example whilst MD5 is a no-no in a security context (you may not use an MD5 digests for a certificate) , you can still use it outside of a security context.  e.g. just checking a file was not corrupted in a transfer is ok, but it is not ok to use md5 to check a file was not tampered with.  
When it comes to encryption you generally will be in a security context (otherwise why bother with encryption), but most of our encryption goes through common code (the secrets or crypto confidential store) which has been well written and uses an allowed algorithm already.

basil: > where you thinking of anything in particular?

https://github.com/jenkinsci/instance-identity-plugin/pull/102

jep/0000/README.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+Some of the cryptography supported in FIPS 140 (especially FIPS 140-2) is considered less secure than some modern alternatives.  FIPS 140-2 cryptography will be used when Jenkins is running in FIPS-140 mode even if that cryptography is considered less secure than other alternatives.
+
+== Infrastructure Requirements
+
+There are no new infrastructure requirements related to this proposal.
+
+== Testing
+
+New code that changes behaviour when it detects the desire to run in FIPS-140 mode would be covered by unit tests like any other code change.  There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity of maintaining this setup.
```
basil, line 114: > There would be no new testing introduced in a FIPS-140 compliant environment due to the complexity of maintaining this setup.

An assertion made without evidence. Why is it true?

Even if maintaining an automated test setup is too complex, the JEP template asks that authors describe their manual test plan. Why is that not done in this JEP?

jep/0000/README.adoc
jtnord: which is not part of Jenkins core.

basil: Are other changes needed outside of Jenkins core? How do we know which plugins are affected? How do we know which plugins are not affected? In general, project planning in Jenkins-land requires one to do searches throughout the plugin ecosystem.

jtnord: > Are other changes needed outside of Jenkins core

some highly used plugins need changes in order to be fips compliant, but they do not need to be changed to run Jenkins in a fips compliant way (which is this jep) 

`git-plugin`, `mina-*plugins` (bouncycastle needs to be updated to use the [new API in addition](https://github.com/jenkinsci/bouncycastle-api-plugin/blob/master/src/main/java/jenkins/bouncycastle/api/BouncyCastlePlugin.java#L26-L28)).  

But these are not required to run Jenkins in a fips compliant nature as they are plugins.

`publish-over-jcifs` would likely need changes to be fps compliant (NTLM auth for example) if anyone cared to make the changes.

> How do we know which plugins are affected?

who is "we" here, the Jenkins community,  the end user?

There is no facility here to record any specific plugin version as FIPS compliant.  This JEP is about making it possible to run Jenkins core in a fips compliant way, not recording all the compliance of 1000+ plugins with N versions.

As for the end user, this is documented https://www.jenkins.io/doc/book/system-administration/FIPS-140/ it is not handled by the community.

Plugins authors can check the guidelines at https://www.jenkins.io/doc/developer/FIPS-140/  but that alone may not be enough, if in doubt they can ask on the mailing list (noted that this point this needs adding to that page)

>  In general, project planning in Jenkins-land requires one to do searches throughout the plugin ecosystem.

see line 69.  this has been explicitly called out of scope.
   "Configuration of the JVM, platform etc. is out-of-scope, as is making every feature/plugin compliant."

Thus checking every single plugin or even the top 500 is out of scope.  If you want to take it upon yourself in your free time to provide this service to the community and to keep it updated for every plugin release then feel free.

If an end user desires to have some fips compliance then they can seek a commercial vendor that has the resources to help them, or they can go it alone and perform their own checks at their own risk.

jep/0000/README.adoc
jtnord: I will create a PR to update this section

basil:
I think any Jenkins community JEP should have its minimum viable product (MVP) defined in a way that benefits at least a small portion of the Jenkins community. While the top 500 plugins is likely too broad a scope, a scope of just "Jenkins core, without documentation of the configuration required to run Jenkins in a FIPS-140 compliant manner" is too narrow. Its benefit to the Jenkins community is minimal to nonexistent. I mean, it is simply absurd to think about running Jenkins core with zero plugins. Core's automated tests do not even attempt to do that. I think the scope of this project should be increased to some reasonable MVP.

jep/0000/README.adoc
basil: ⇨ #404

PR #401, opened 2023-10-31
Improve template

basil:
Fixes several issues I have had with the template for a long time:

- Move the **Specification** section to after the **Reasoning** section. The current order has always seemed backwards to me. The **Specification** section essentially covers the implementation details of the solution. How is this to be understood by the reader if that reader does not first understand the problem and its high-level solution, which are covered in the **Motivation** and **Reasoning** sections? By way of analogy, when writing a mathematical proof, the proof *ends* with the thing you're trying to prove; it doesn't start with it. A mathematical proof starts with definitions, then proceeds to propositions, lemmas, corollaries, and ultimately the explanation of why a statement is true (the proof). This PR simply applies the same logical structure to this document.
- Add clarity to the **Motivation** section based on personal experience. One of the most common mistakes people make in this part of the design process is to write the problem in terms of the solution, so explicitly call out this trap in the hints.
- Add clarity to the **Reasoning** section based on personal experience. One of the most common mistakes people make in this part of the design process is to get bogged down in implementation details or not describing other possible designs, so explicitly call out these traps in the hints.
- Add clarity to the **Testing** section to reflect that testing not just *can* be done but *will* be done. We do not accept untested changes.

### Testing done

Read the new text and it makes sense to me. I have been using these ideas for years in other roles and communities.

### DIFF
--- a/jep-template/0000/README.adoc
+++ b/jep-template/0000/README.adoc
@@ -94,19 +94,6 @@ Give a short (200 word) description of the technical issue addressed.
 * Do not talk about history or why this needs to be done. Instead, add the history to the Motivation section.
 ====
 
-== Specification
-
-[TIP]
-====
-Provide a detailed specification of what is being proposed.
-Be as technical and detailed as needed to allow new or existing Jenkins developers
-to reasonably understand the scope/impact of an implementation.
-
-* Use present tense - describe what the proposal "does" (as if it were already done), not what it will do.
-* Do not discuss alternative designs that were rejected - those belong in the Reasoning section.
-* Avoid in-depth discussion or justification of design choices - that belongs in the Reasoning section.
-====
-
 == Motivation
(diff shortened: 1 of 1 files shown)


daniel-beck (approved):
Unsure whether my vote counts, not being an editor, but this looks great.

Wrt section order, I remembered conversations like https://github.com/jenkinsci/jep/pull/1#discussion_r138728754 about JEP-1, but looking at [PEP-1](https://peps.python.org/pep-0001/#pep-formats-and-templates) and the [template in PEP-12](https://peps.python.org/pep-0012/#suggested-sections), they all use the ordering being introduced here. So I'm unsure why we ended up with the current order.

MarkEWaite (approved):
Thanks very much.  I agree with @daniel-beck that this is a good change.

PR #402, opened 2023-10-31
Add a work estimate section to the template

basil:
The Jenkins project does not have dedicated project managers, so architects ought to consider the project management implications of their proposals. My personal experience has been that considering these implications at the time of project design can significantly improve the resulting design by identifying task dependency bottlenecks (and forcing the author to think about if and how they can be resolved) as well as staffing requirements (and forcing the author to revise the design if it is impractical within the realities of staffing constraints). Most importantly, the act of breaking down a prototype into the tasks needed to productize it serves as a forcing function to ensure that the prototype is complete enough to serve its purpose of informing a design. If a prototype is so preliminary that one cannot clearly envision the tasks needed to complete it, then that is an indication that the prototype has failed to explore all the relevant issues in enough depth and that the design is likely to contain gaps and/or errors in judgment.

### Testing done

I have been using these ideas for years in other roles and communities.

### DIFF
--- a/jep-template/0000/README.adoc
+++ b/jep-template/0000/README.adoc
@@ -203,6 +203,25 @@ but must be completed before any JEP is given
 JEPs which will not include code changes may omit this section.
 ====
 
+== Work Estimates
+
+[TIP]
+====
+Provide a clear description of the high-level tasks needed to productize the prototype implementation.
+Ideally these tasks would be turned into work items in an issue tracking system (ITS) like Jira or GitHub Issues.
+As a general rule, it should be possible to complete each task within 1-3 days;
+if a task takes a week or longer, it has not been broken down with enough granularity.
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
I think this is a good addition to the JEP process.  @timja as the other active JEP editor, are you OK with the addition?

I'll adapt #400 to include the recent JEP improvements

timja (approved):

lemeurherve:
FTR, I've also started working on a JEP for https://github.com/jenkins-infra/helpdesk/issues/2649 including this section.

timja:
shouldn't that be an IEP? https://github.com/jenkins-infra/iep

lemeurherve:
@timja: after discussing it with @dduportal and as we'd like to establish a contract with Core to follow UC redirections to mirrors, we think the scope of this project is more large than an IEP. This can be challenged though.

